PR #32, opened 2024-04-17
CI, formatting, linting

surgura:
- Added [dev] optional to package (pip install ".[dev]")
- Added tools formatting & linting to the [dev] optional
- Added script that runs these tools and applies fixes whenever able
- Added CI that checks code using these tools
- Adjusted code to pass checks (no functional changes)

Don't pay too much attention to the docstrings. They are mostly to have the CI pass and we will make them better when we remodel the code structure.

### DIFF
--- a/.github/workflows/codetools.yml
+++ b/.github/workflows/codetools.yml
@@ -1,0 +1,33 @@
+name: ci
+
+on: [push, pull_request]
+
+env:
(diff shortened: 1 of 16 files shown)


surgura:
I forgot to rebaseto dev, let me do that

erikvansebille (approved):

PR #36, opened 2024-06-05
Ship underwater st test

surgura:
- Implemented test for ship underwater st. Other tests will follow, but let's review this one so we agree on how to test.
- Refactored and made ship underwater st work so it passes the test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,7 +21,7 @@ def change_test_dir(request, monkeypatch):
 @pytest.fixture
 def tmp_dir_factory() -> Generator[Callable[[str], str], None, None]:
     """
-    Return a function that can generate a random directory and returns its path.
+    Yields a function that can generate a random directory and returns its path.
 
     Created directories are automatically deleted after the test.
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,6 +1,9 @@
 """Test configuration that is ran for every test."""
 
 import pytest
+import tempfile
+import shutil
+from typing import Generator, Callable
 
 
 @pytest.fixture(autouse=True)
@@ -12,3 +15,23 @@ def change_test_dir(request, monkeypatch):
     :param monkeypatch: -
     """
     monkeypatch.chdir(request.fspath.dirname)
+
+
+@pytest.fixture
+def tmp_dir_factory() -> Generator[Callable[[str], str], None, None]:
+    """
+    Returns a functions that can generate a random directory and returns its path.
+
+    Created directories are automatically deleted after the rest.
+    """
+    created_dirs = []
+
+    def _create_temp_dir(suffix: str):
+        dir_path = tempfile.mkdtemp(suffix=suffix)
+        created_dirs.append(dir_path)
+        return dir_path
+
+    yield _create_temp_dir
+
+    for dir_path in created_dirs:
+        shutil.rmtree(dir_path)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,9 +1,10 @@
 """Test configuration that is ran for every test."""
 
-import pytest
-import tempfile
 import shutil
-from typing import Generator, Callable
+import tempfile
+from typing import Callable, Generator
+
+import pytest
 
 
 @pytest.fixture(autouse=True)
@@ -20,9 +21,11 @@ def change_test_dir(request, monkeypatch):
 @pytest.fixture
 def tmp_dir_factory() -> Generator[Callable[[str], str], None, None]:
     """
-    Returns a functions that can generate a random directory and returns its path.
+    Return a function that can generate a random directory and returns its path.
 
     Created directories are automatically deleted after the rest.
+
+    :yields: A generator for temporary directories that will be automatically removed.
     """
     created_dirs = []
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -21,7 +21,7 @@ def change_test_dir(request, monkeypatch):
 @pytest.fixture
 def tmp_dir_factory() -> Generator[Callable[[str], str], None, None]:
     """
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -12,3 +16,25 @@ def change_test_dir(request, monkeypatch):
     :param monkeypatch: -
     """
     monkeypatch.chdir(request.fspath.dirname)
+
+
+@pytest.fixture
+def tmp_dir_factory() -> Generator[Callable[[str], str], None, None]:
```
erikvansebille, line 22: For my information/learning; why do you use a custom `tmp_dir_factory` function? Why not use the built-in [`tmpdir` of pytest](https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture)? That is what we do in `parcels` (e.g. [here](https://github.com/OceanParcels/parcels/blob/3652e364d2a424ca0ddeea3109ddb5abad91d6d1/tests/test_particlefile.py#L41))

tests/instruments/test_ship_underwater_st.py
```diff
@@ -1,29 +1,102 @@
...
+    for i, (obs_i, exp) in enumerate(
+        zip(results.sel(trajectory=traj).obs, expected_obs, strict=True)
+    ):
+        obs = results.sel(trajectory=traj, obs=obs_i)
+        for var in variables:
+            obs_value = obs[var].values.item()
+            exp_value = exp[var]
+            assert np.isclose(
+                obs_value, exp_value
+            ), f"Observation incorrect {i=} {var=} {obs_value=} {exp_value=}."
```
erikvansebille, line 102: Would it make more sense to print `obs_i`?
```suggestion
            ), f"Observation incorrect {obs_i=} {var=} {obs_value=} {exp_value=}."
```

virtual_ship/instruments/ship_underwater_st.py
```diff
@@ -1,11 +1,14 @@
...
 import numpy as np
-from parcels import FieldSet, JITParticle, ParticleSet, Variable
+from parcels import FieldSet, ParticleSet, ScipyParticle, Variable
 
 from ..spacetime import Spacetime
 
-_ShipSTParticle = JITParticle.add_variables(
+# we specifically use ScipyParticle because we have many small calls to execute
+# JITParticle would require compilation every time
+# this ends up being faster
```
erikvansebille, line 10: Interesting. Parcels should only recompile if the kernels changed, so apparently there's an issue there? Or is it just that the overhead of loading the C-libraries etc makes JIT slower even without recompilation?

virtual_ship/instruments/ship_underwater_st.py
```diff
@@ -42,26 +49,31 @@ def simulate_ship_underwater_st(
...
         lat=0.0,
         depth=depth,
         time=0,  # same for time
     )
 
     # define output file for the simulation
-    out_file = particleset.ParticleFile(
-        name=out_file_name,
-    )
+    # the default outputdt is good(infinite), as we want to just want to write at the end of every call to 'execute'
```
erikvansebille, line 59: Still, I would then explicitly set `outputdt=np.inf`, as the default may(?) change at some point in parcels?

virtual_ship/instruments/ship_underwater_st.py
```diff
@@ -42,26 +49,31 @@ def simulate_ship_underwater_st(
...
+    out_file = particleset.ParticleFile(name=out_file_name)
 
+    # iterate over each points, manually set lat lon time, then
```
erikvansebille, line 62:
```suggestion
    # iterate over each point, manually set lat lon time, then
```

virtual_ship/instruments/ship_underwater_st.py
```diff
@@ -42,26 +49,31 @@ def simulate_ship_underwater_st(
...
+    # execute the particle set for one step, performing one set of measurement
     for point in sample_points:
         particleset.lon_nextloop[:] = point.location.lon
         particleset.lat_nextloop[:] = point.location.lat
-        particleset.time_nextloop[:] = point.time
+        particleset.time_nextloop[:] = fieldset.time_origin.reltime(
+            np.datetime64(point.time)
```
erikvansebille, line 68: Would it make sense to use `np.datetime64` all throughout VirtualShip? That is what we tend to use in `parcels` too. Or would you prefer to keep `datetime`?

erikvansebille (commented):
Looks like a good and useful test flow. See some comments below

tests/conftest.py
surgura: I was unaware of its existance and have replaced

tests/instruments/test_ship_underwater_st.py
surgura: correct, adjusted and enumerate removed

virtual_ship/instruments/ship_underwater_st.py
surgura: Right. I assumed it was the recompilation, but maybe its something else related. Does parcels cache the compiled kernel even if you call 'execute' multiple times?

virtual_ship/instruments/ship_underwater_st.py
surgura: fair enough. adjusted

virtual_ship/instruments/ship_underwater_st.py
surgura: to me it is all the same. I tried to remain consistent with datetime, but we can swap to np.datetime64.

erikvansebille (approved):

virtual_ship/instruments/ship_underwater_st.py
surgura: I'll change the comment a bit

PR #37, opened 2024-06-19
Other instrument tests

surgura:
- ADCP instrument cleanup and test
- Ship ST instrument test improvements

### DIFF
--- a/tests/instruments/test_ship_underwater_st.py
+++ b/tests/instruments/test_ship_underwater_st.py
@@ -83,7 +83,7 @@ def test_simulate_ship_underwater_st(tmpdir: py.path.LocalPath) -> None:
     results = xr.open_zarr(out_path)
 
     # below we assert if output makes sense
-    assert len(results.trajectory) == 1  # expect a singe trajectory
+    assert len(results.trajectory) == 1  # expect a single trajectory
     traj = results.trajectory.item()
     assert len(results.sel(trajectory=traj).obs) == len(
         sample_points
--- a/tests/instruments/test_ship_underwater_st.py
+++ b/tests/instruments/test_ship_underwater_st.py
@@ -23,8 +23,8 @@ def test_simulate_ship_underwater_st(tmpdir: py.path.LocalPath) -> None:
 
     # where to sample
     sample_points = [
-        Spacetime(Location(0, 0), base_time + datetime.timedelta(seconds=0)),
-        Spacetime(Location(1, 0), base_time + datetime.timedelta(seconds=1)),
+        Spacetime(Location(0, 1), base_time + datetime.timedelta(seconds=0)),
+        Spacetime(Location(2, 3), base_time + datetime.timedelta(seconds=1)),
     ]
 
     # expected observations at sample points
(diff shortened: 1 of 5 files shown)


tests/instruments/test_ship_underwater_st.py
```diff
@@ -80,21 +81,23 @@ def test_simulate_ship_underwater_st(tmpdir: py.path.LocalPath) -> None:
...
-    assert len(results.trajectory) == 1  # expect a singe trajectory
+    assert len(results.trajectory) == 1  # expect a single trajectory
     traj = results.trajectory.item()
     assert len(results.sel(trajectory=traj).obs) == len(
         sample_points
     )  # expect as many obs as sample points
 
     # for every obs, check if the variables match the expected observations
-    for obs_i, exp in zip(results.sel(trajectory=traj).obs, expected_obs, strict=True):
+    for i, (obs_i, exp) in enumerate(
```
surgura, line 94: explicitly iterate over i, because printing obs_i in case of failed assert turned out to not be useful, and with i we can tell which observation failed.

virtual_ship/location.py
```diff
@@ -7,23 +7,23 @@
 class Location:
```
surgura, line 7: just swapping y,x for x,y. changed the rest of the code accordingly

virtual_ship/sailship.py
```diff
@@ -203,10 +203,11 @@ def sailship(config: VirtualShipConfiguration):
     print("Simulating onboard ADCP.")
     simulate_adcp(
         fieldset=config.adcp_fieldset,
-        out_file_name=os.path.join("results", "adcp.zarr"),
+        out_path=os.path.join("results", "adcp.zarr"),
         max_depth=config.ADCP_settings["max_depth"],
         min_depth=-5,
-        bin_size=config.ADCP_settings["bin_size_m"],
+        num_bins=(-5 - config.ADCP_settings["max_depth"])
```
surgura, line 209: so, before the argument was bin_size. I did not like that because it's very depending on the min and max. now you just choose the number of bins, and it doesn't break if you change min/max

surgura: to keep the behaviour in sailship the same we apply this small conversion formula. for now until we rework the config file

virtual_ship/instruments/adcp.py
```diff
@@ -53,14 +56,22 @@ def simulate_adcp(
     )
 
     # define output file for the simulation
-    out_file = particleset.ParticleFile(
-        name=out_file_name,
-    )
+    # outputdt set to infinie as we want to just want to write at the end of every call to 'execute'
```
erikvansebille, line 59:
```suggestion
    # outputdt set to infinite as we want to just want to write at the end of every call to 'execute'
```

virtual_ship/location.py
erikvansebille: Hmm, while it's often more common to give locations in (lon, lat), there is also a point to be made to keep the (lat, lon) order, as Parcels for examples samples like (time, depth, lat, lon). @ammedd, what do you think?

virtual_ship/sailship.py
erikvansebille: @ammedd, what is the setting on an actual ADCP? I think we should keep as close as possible to what is done in the real instruments

ammedd: This has been on my Wishlist (https://github.com/OceanParcels/Virtual_ship_classroom/issues/24) and is more realistic, so go ahead

virtual_ship/location.py
ammedd: Lat, lon. It is used everywhere with python (and already confusing enough as it is)

surgura: Ok, ordered it lat lon, y x, V U everywhere

erikvansebille (approved):

PR #38, opened 2024-06-27
Ctd cleanup and test

surgura:
* Cleaned up CTD
* Wrote test for CTD

Fields varying with time are not tested, as it's quite tricky to predict the expected observations.

There is also an issue currently, where the simulation always goes over the complete fieldset. Have to check if we can tall Parcels to stop simulating when all particles are done, and maybe skip over parts where no particles have been deployed yet.

### DIFF
--- a/virtual_ship/instruments/ctd.py
+++ b/virtual_ship/instruments/ctd.py
@@ -75,7 +75,7 @@ def simulate_ctd(
     fieldset_endtime = fieldset.time_origin.fulltime(fieldset.U.grid.time_full[-1])
 
     # deploy time for all ctds should be later than fieldset start time
-    if not all([ctd.spacetime.time <= fieldset_starttime for ctd in ctds]):
+    if not all([ctd.spacetime.time >= fieldset_starttime for ctd in ctds]):
         raise ValueError("CTD deployed before fieldset starts.")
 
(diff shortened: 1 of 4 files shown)


virtual_ship/instruments/ctd.py
```diff
@@ -37,77 +40,89 @@ def _sample_salinity(particle, fieldset, time):
...
+    :param out_path: The path to write the results to.
+    :param ctds: A list of CTDs to simulate.
     :param outputdt: Interval which dictates the update frequency of file output during simulation
+    :raises ValueError: Whenever provided CTDs, fieldset, are not compatible with this function.
     """
-    lon = [ctd.spacetime.location.lon for ctd in ctds]
-    lat = [ctd.spacetime.location.lat for ctd in ctds]
-    time = [ctd.spacetime.time for ctd in ctds]
+    WINCH_SPEED = 1.0  # sink and rise speed in m/s
+    DT = 10.0  # dt of CTD simulation integrator
```
erikvansebille, line 72: Why is this variable in capitals? We normally use lowercase for `dt`

surgura: it is a constant, and in PEP8 constants have to be all caps

PR #39, opened 2024-07-03
Drifter test

surgura:
* Cleaned up Drifters
* Wrote test for Drifters


virtual_ship/spacetime.py
```diff
@@ -7,7 +7,6 @@
 
 
 @dataclass
-# TODO I take suggestions for a better name
```
erikvansebille, line 10: I think LocationTime would already be better (less StarTrek ;-) )

erikvansebille (approved):

virtual_ship/spacetime.py
surgura: let me do this in another pr

PR #41, opened 2024-07-16
Config revision

surgura:
* Added typing and initialization in python instead of from JSON.
* Added 'verify' function that somewhat performs the same verification as before.
* Some general cleanup and simplification in sailship function

if you want JSON serialization let me know


virtual_ship/virtual_ship_configuration.py
```diff
@@ -1,173 +1,156 @@
...
+            raise ValueError("Argo float drift depth must be negative or zero.")
+
+        if self.argo_float_config.vertical_speed >= 0:
+            raise ValueError("Argo float vertical speed must be negative.")
+
+        if self.argo_float_config.cycle_days <= 0:
+            raise ValueError("Argo float cycle days must be large than one.")
+
+        if self.argo_float_config.drift_days <= 0:
+            raise ValueError("Argo drift cycle days must be large than one.")
```
erikvansebille, line 144: These don't need to be integer days, right? Could also be fractions of days (to support drifts in hours)?

virtual_ship/virtual_ship_configuration.py
```diff
@@ -1,173 +1,156 @@
...
+        if self.adcp_config.max_depth > 0:
+            raise ValueError("ADCP max depth must be negative.")
+
+    @staticmethod
+    def _is_valid_location(location: Location) -> bool:
+        return (
+            location.lat >= -90
+            and location.lat <= 90
+            and location.lon >= -180
+            and location.lon <= 180
```
erikvansebille, line 155: Hmm, this may be too strict. Many people (including myself) are used to longitudes between 0 and 360 degrees

virtual_ship/virtual_ship_configuration.py
surgura: made them float

virtual_ship/virtual_ship_configuration.py
```diff
@@ -1,173 +1,124 @@
...
+            raise ValueError("Argo float max depth must be negative or zero.")
+
+        if self.argo_float_config.drift_depth > 0:
+            raise ValueError("Argo float drift depth must be negative or zero.")
+
+        if self.argo_float_config.vertical_speed >= 0:
+            raise ValueError("Argo float vertical speed must be negative.")
+
+        if self.argo_float_config.cycle_days <= 0:
+            raise ValueError("Argo float cycle days must be larger than one.")
```
erikvansebille, line 118:
```suggestion
            raise ValueError("Argo float cycle days must be larger than zero.")
```

virtual_ship/virtual_ship_configuration.py
```diff
@@ -1,173 +1,124 @@
...
+
+        if self.argo_float_config.drift_days <= 0:
+            raise ValueError("Argo drift cycle days must be larger than one.")
```
erikvansebille, line 121:
```suggestion
            raise ValueError("Argo drift cycle days must be larger than zero.")
```

virtual_ship/virtual_ship_configuration.py
```diff
@@ -1,173 +1,124 @@
...
-        west = self.region_of_interest["West"]
-        poly = Polygon([(west, north), (west, south), (east, south), (east, north)])
-        # validate input, raise ValueErrors if invalid
-        if (
-            self.region_of_interest["North"] > 90
-            or self.region_of_interest["South"] < -90
-            or self.region_of_interest["East"] > 180
-            or self.region_of_interest["West"] < -180
-        ):
-            raise ValueError("Invalid coordinates in region of interest")
```
erikvansebille, line 66: Well, I think the test that latitude is between -90 and 90 is stil valid, so I'd keep that in. And we can test that longitude is either between -180 and 180, or between 0 and 360. In my last comment, I didn't mean to propose to complete remove these checks

surgura: I am not sure if that makes sense, because we cannot deduce what range the user wants to use. Functionally it does not matter, and such checks might only annoy the user? I guess we could check for larger than -180 and smaller than 360, but I'm not sure if that is useful. It's a bit odd if a user provides 720, but it's technically fine?

erikvansebille (approved):

PR #46, opened 2024-07-30
Sail by waypoint

surgura:


virtual_ship/instruments/ctd.py
```diff
@@ -71,6 +71,13 @@ def simulate_ctd(
     WINCH_SPEED = 1.0  # sink and rise speed in m/s
     DT = 10.0  # dt of CTD simulation integrator
 
+    if len(ctds) == 0:
+        print(
+            "No CTDs provided. Parcels currently crashes when providing an empty particle set, so no CTD simulation will be done and no files will be created."
+        )
+        # TODO when parcels supports it this check can be removed.
```
erikvansebille, line 78: OK, so this is what https://github.com/OceanParcels/parcels/issues/1629 is about? I'll try fix that asap

virtual_ship/planning_error.py
```diff
@@ -0,0 +1,7 @@
+"""PlanningError Exception."""
+
+
+class PlanningError(RuntimeError):
+    """An error when checking the schedule or during sailing."""
```
erikvansebille, line 5: Can this be any more specific? Provide some further info when/where/why the error was caused. Otherwise, we'll have to do a lot of debugging for students...

virtual_ship/sailship.py
```diff
@@ -3,185 +3,107 @@
...
-
-        # find argo float deployments to be done at this location
-        argos_here = set(
-            [
-                argo
-                for argo in argo_locations - argo_locations_visited
-                if all(
-                    np.isclose([argo.lat, argo.lon], [route_point.lat, route_point.lon])
+                # use 20 minutes to case the CTD
+                time += timedelta(minutes=20)
```
erikvansebille, line 64: I hadn't seen this before. Does it make sense to have this hardcoded? Can confuse users who think their planning fits, but who don't realise another 20 minutes is added for each CTD. Perhaps add a warning/notice? @ammedd, what do you think?

virtual_ship/sailship.py
```diff
@@ -3,185 +3,107 @@
...
+            case InstrumentType.DRIFTER:
+                drifters.append(
+                    Drifter(
+                        spacetime=spacetime,
+                        depth=-config.drifter_fieldset.U.depth[0],
```
erikvansebille, line 69: What if the user wants to sample at a different depth? E.g. at 15m (SVP-style) instead of 0m?

virtual_ship/sailship.py
```diff
@@ -229,61 +151,55 @@ def sailship(config: VirtualShipConfig):
...
-    return sample_points
+        time_to_reach = timedelta(seconds=distance / ship_speed)
+        arrival_time = time + time_to_reach
+
+        if wp_next.time is None:
+            time = arrival_time
+        elif arrival_time > wp_next.time:
+            raise PlanningError(
+                "Waypoint planning is not valid: would arrive too late a waypoint."
+            )
```
erikvansebille, line 203: Give a bit more context? Which point would be arrived too late? This helps debugging schedules later

virtual_ship/waypoint.py
```diff
@@ -0,0 +1,16 @@
...
+from dataclasses import dataclass
+from datetime import datetime
+
+from .instrument_type import InstrumentType
+from .location import Location
+
+
+@dataclass
+class Waypoint:
+    """A Waypoint to sail to."""
```
erikvansebille, line 12:
```suggestion
    """A Waypoint to sail to and (optionally) deploy an instrument at."""
```

virtual_ship/sailship.py
surgura: it is a parameter in the config now

virtual_ship/sailship.py
surgura: Moved this to config

virtual_ship/priority_queue.py
erikvansebille, line 1: Did you write this file specifically for the VSC? Seems fairly generic, would it not better fit in a `tools` folder or so? Or is there a library/package that provides this itself? I understand the code itself is not too complicated, but better to use tested packages for generic algorithms like these?

virtual_ship/sailship.py
```diff
@@ -1,219 +1,83 @@
...
 from .spacetime import Spacetime
-from .virtual_ship_configuration import VirtualShipConfig
+from .virtual_ship_config import VirtualShipConfig
+from .waypoint import Waypoint
 
 
 def sailship(config: VirtualShipConfig):
     """
-    Use parcels to simulate the ship, take ctd_instruments and measure ADCP and underwaydata.
+    Use parcels to simulate a cruise.
```
erikvansebille, line 30: I really don't like the word 'cruise'. I know it's common parlour in the oceanographic community; but can we please change to 'virtual ship expedition' or simply 'expedition' throughout?

erikvansebille, line 30: We are changing the name of the code to Parcels with capital P in normal text (like NumPy) 
```suggestion
    Use Parcels to simulate a cruise.
```

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
 
-def shiproute(config: VirtualShipConfig, dt: timedelta) -> list[Location]:
+def _simulate_schedule(
+    waypoints: list[Waypoint],
+    projection: pyproj.Geod,
+    config: VirtualShipConfig,
+) -> _ScheduleResults:
     """
-    Take in route coordinates and return lat and lon points within region of interest to sample.
+    Simulate the sailing and aggregate what measurements should be simulated.
```
erikvansebille, line 110:
```suggestion
    Simulate the sailing and aggregate the virtual measurements that should be taken.
```

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
 
+    :param waypoints: The schedule.
```
erikvansebille, line 112: A list of waypoint objects?

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
+    :param projection: Projection used to sail between waypoints.
```
erikvansebille, line 113: Also mention what the default is?

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
     :param config: The cruise configuration.
-    :param dt: Sailing time between each discrete route point.
-    :returns: lat and lon points within region of interest to sample.
+    :returns: Results from the simulation.
+    :raises NotImplementedError: When not supported instruments are encountered.
```
erikvansebille, line 116:
```suggestion
    :raises NotImplementedError: When unsupported instruments are encountered.
```

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
         )
     )
 
-    return sample_points
+
+def _verify_waypoints(
+    waypoints: list[Waypoint], ship_speed: float, projection: pyproj.Geod
+) -> None:
+    """
+    Verify waypoints are ordered by time, first waypoint has a start time, and that schedule is feasible in terms of time if no unexpected things happen.
```
erikvansebille, line 389:
```suggestion
    Verify waypoints are ordered by time, first waypoint has a start time, and that schedule is feasible in terms of time if no unexpected events happen.
```

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
+        [
+            next.time >= cur.time
+            for cur, next in zip(timed_waypoints, timed_waypoints[1:])
+        ]
+    ):
+        raise PlanningError(
+            "Each waypoint should be timed after all previous waypoints"
+        )
+
+    # check that ship will arrive on time at each waypoint (in case nothing goes wrong)
```
erikvansebille, line 412:
```suggestion
    # check that ship will arrive on time at each waypoint (in case no unexpected events happen)
```

erikvansebille: Let's call these 'unexpected events' throughout, so we can later add storms etc if we want

virtual_ship/sailship.py
```diff
@@ -222,68 +86,348 @@ def sailship(config: VirtualShipConfig):
...
+        distance = geodinv[2]
+
+        time_to_reach = timedelta(seconds=distance / ship_speed)
+        arrival_time = time + time_to_reach
+
+        if wp_next.time is None:
+            time = arrival_time
+        elif arrival_time > wp_next.time:
+            raise PlanningError(
+                f"Waypoint planning is not valid: would arrive too late at a waypoint requiring time {wp_next.time}"
```
erikvansebille, line 430: Can this error also provide _which_ waypoint is invalid? That will help debugging

virtual_ship/priority_queue.py
surgura: I don't disagree, but I could not find a reliable package that provides a priority queue. Python itself does not provide a priority queue in the standard library either. There is queue.PriorityQueue, but it is designed for multithreading situations. We could use it for this single threaded situation, but the way we would use its interface would in my opinion be quite unclear to the reader. Python provides heapq, which pop and pushes to a list to create a priority queue. I opted to use that, but to make it more readable I created this wrapper class; it does not do anything but make heapq more readable really

erikvansebille: OK thanks for the clarification; I'm fine with this then

surgura: Ok I found a library, sortedcontainers.SortedList. This has an acceptable interface and is reasonably well-maintained. I have replaced our wrapper class with that library

virtual_ship/sailship.py
surgura: I am not sure what you mean? this function does not provide a default?

virtual_ship/sailship.py
surgura: is this better?

```
virtual_ship.planning_error.PlanningError: Waypoint planning is not valid: would arrive too late at a waypoint number 1. location: Location(latitude=-43.081289, longitude=63.743631) time: None instrument: InstrumentType.CTD
```

virtual_ship/sailship.py
erikvansebille: Well, I think the default is `pyproj.Geod(ellps="WGS84")`? Mention that here in the docstring?

erikvansebille (approved):

virtual_ship/sailship.py
surgura: This function itself does not provide a default. The projection is passed when called the function. You might be referring to the sailship function, which indeed uses that projection.

PR #48, opened 2024-08-07
verify waypoints are on water

### LINKED ISSUE #25: Ship and measurements on land
Currently the virtual ship and can "sail" on land and do CTD measurements/deployments...

A possible way to fix this is using status codes in the kernel loops
https://github.com/OceanParcels/parcels/blob/daf05b71174ba2c687bfe0468ed27e1e64f17d06/docs/examples/tutorial_kernelloop.ipynb#L300
Or check using the CMEMS input data? Or...

surgura:
verified when checking the waypoints, not during simulation

closes #25

### DIFF
--- a/virtual_ship/sailship.py
+++ b/virtual_ship/sailship.py
@@ -36,7 +36,9 @@ def sailship(config: VirtualShipConfig):
     # projection used to sail between waypoints
     projection = pyproj.Geod(ellps="WGS84")
 
-    _verify_waypoints(config.waypoints, config.ship_speed, projection=projection)
+    _verify_waypoints(
+        config.waypoints, config.ship_speed, projection=projection, config=config
+    )
 
     # simulate the sailing and aggregate what measurements should be simulated
     schedule_results = _simulate_schedule(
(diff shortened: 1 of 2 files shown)


virtual_ship/sailship.py
```diff
@@ -412,6 +419,47 @@ def _verify_waypoints(
...
+    fieldset = available_fieldsets[0]
+    # get waypoints with 0 UV
+    land_waypoints = [
+        (wp_i, wp)
+        for wp_i, wp in enumerate(waypoints)
+        if fieldset.UV.eval(
+            0, 0, wp.location.lat, wp.location.lon, applyConversion=False
+        )
+        == (0.0, 0.0)
+    ]
```
erikvansebille, line 456: Make this a more general method, so that someone could later change the way in which landpoints are defined (e.g. through a mask-array, instead of zero-velocity condition)

virtual_ship/sailship.py
```diff
@@ -412,6 +419,47 @@ def _verify_waypoints(
...
+    # raise an error if there are any
+    if len(land_waypoints) > 0:
+        raise PlanningError(
+            f"The following waypoints are on land: {['#' + str(wp_i) + ' ' + str(wp) for (wp_i, wp) in land_waypoints]}"
+        )
```
erikvansebille, line 461: Also create a unit test to check that this `PlanningError` is indeed raised when expected?

virtual_ship/sailship.py
surgura: see update

virtual_ship/sailship.py
surgura: added test

erikvansebille (approved):

PR #49, opened 2024-08-07
move lat lon test into location class

### LINKED ISSUE #47: Location check in location class
Currently it's in virtual_ship_config. Move that to the location class itself.

surgura:
closes #47


erikvansebille (approved):
Nice clean implementation!

Issue #30, opened 2024-04-16
Add noise to "measured data"


ammedd:
Some random noise is currently added to the CTD data in the postprocessing. Please review this and consider to add noise to all other Temperature and Salinity measurements.

surgura:
Starting with adding noise to CTD, other instruments come after

PR #50, opened 2024-08-10
CTD noise and CNV output format

surgura:
#30 for CTD only


surgura:
Sorry for the one million files in the test zarr. I could also generate that on the fly instead of saving? maybe that's beter, not sure

erikvansebille:
Yes, I think that creating the zarr on the fly is better; as it will then also check whether zarr files are created as expected. Alternative is to put .zarr files in the gitignore?

Oh, and should the `CNV` is the title of this PR not be `CSV`?

surgura:
> Yes, I think that creating the zarr on the fly is better; as it will then also check whether zarr files are created as expected. Alternative is to put .zarr files in the gitignore?
> 
> Oh, and should the `CNV` is the title of this PR not be `CSV`?

The file format is CNV, a file format created by Seabird(the microcat brand)
The zarr file is in the test directory and used for the test. There already is a test for the creation of the zarr so that is not needed. I can either create it before running the rest by using the ctd function, but I think it might be better to replace this huge file with a very short zarr. That way a we have a fast test without having too much data on git

erikvansebille:
By changing the `chunks` parameter in the ParticleFile object creation, you can also limit the number of files (=chunks) generated. See the note at https://docs.oceanparcels.org/en/latest/examples/tutorial_parcels_structure.html#4.-Execution-and-output

surgura:
> By changing the `chunks` parameter in the ParticleFile object creation, you can also limit the number of files (=chunks) generated. See the note at https://docs.oceanparcels.org/en/latest/examples/tutorial_parcels_structure.html#4.-Execution-and-output

Ok makes sense I will keep that in mind. I have create a very small one now and I think this should be fine as test input?

tests/instruments/test_ctd.py
```diff
@@ -33,7 +33,7 @@ def test_simulate_ctds(tmpdir: py.path.LocalPath) -> None:
         CTD(
             spacetime=Spacetime(
                 location=Location(latitude=1, longitude=0),
-                time=base_time + datetime.timedelta(minutes=5),
+                time=base_time + datetime.timedelta(hours=0),
```
erikvansebille, line 36: Why not simply as below?
```suggestion
                time=base_time,
```

virtual_ship/make_realistic/ctd_make_realistic.py
```diff
@@ -0,0 +1,187 @@
...
+        * np.maximum(1.0 + depth / 1000.0, 0.1)
+        * (
+            np.array([opensimplex.noise2(n / 30, 0.0) for n in range(len(temperature))])
+            ** 8
+        )
+    )
+    opensimplex.seed(np.random.randint(999999))
+    noise3 = 0.1 * (
+        np.array([opensimplex.noise2(n / 8, 0.0) for n in range(len(temperature))]) ** 2
+    )
```
erikvansebille, line 89: Should we provide a bit of comment/background why these specific values are used for the noise? Do they come from the  CTD manual, as the typical instrument noise?

virtual_ship/make_realistic/ctd_make_realistic.py
```diff
@@ -0,0 +1,187 @@
...
+        * np.maximum(1.0 + depth / 1000.0, 0.1)
+        * np.random.random_sample(len(temperature))
+    )
+    drift = (
+        0.3
+        * (np.random.random() * 2.0 - 1)
+        * np.array(range(len(temperature)))
+        / len(temperature)
+    )
+    opensimplex.seed(np.random.randint(999999))
```
erikvansebille, line 77: Why is the seed deteministic? For reproducibility, I guess? But would it not be more realistic to have noise with a random (clock-based?) seed?

virtual_ship/make_realistic/ctd_make_realistic.py
```diff
@@ -0,0 +1,187 @@
...
+    noise3 = 0.1 * (
+        np.array([opensimplex.noise2(n / 8, 0.0) for n in range(len(temperature))]) ** 2
+    )
+
+    return temperature + surface_noise + drift + noise + noise3 + noise2
+
+
+def _add_salinity_noise(salinity: np.ndarray, depth: np.ndarray) -> np.ndarray:
```
erikvansebille, line 94: Some comments/questions as with the temperature noise function

virtual_ship/make_realistic/ctd_make_realistic.py
```diff
@@ -0,0 +1,187 @@
...
+    :param temperatures: Temperature data.
+    :param depths: Depth data.
+    :param salinities: Salinity data.
+    :returns: The CNV.
+
+    """
+    header = rf"""
+* Sea-Bird SBE 9 Data File:
+* FileName = {filename}
+# start_time = Aug 02 2024 21:29:45 [NMEA time, first data scan]
```
erikvansebille, line 151: Should this starttime not be taken from the virtual ship cast time?

virtual_ship/make_realistic/ctd_make_realistic.py
surgura: I have added docstrings to the function. Basically I tried to add arbitrary nonphysical noise to the data so it looks somewhat like real CTD plots we get on the ship here. Emma will have another look another time to improve it, but we decided that for now this is good enough.

virtual_ship/make_realistic/ctd_make_realistic.py
surgura: It is not deterministic :) I seed with a random number created by the random module. The opensimplex library is a bit annoying so I chose to seed it like this. I think we will use something different once we revisit the noise, but for now I think this is fine.

virtual_ship/make_realistic/ctd_make_realistic.py
surgura: Done

erikvansebille (approved):

Issue #53, opened 2024-08-16
interactive loop


surgura:
Create an interactive loop that asks users to make decisions each time something unexpected happens. The remainder of the schedule including the first waypoint that fails is provided. The user has to input a new schedule to continue. At each decision point the software creates a checkpoint so the user can continue even when quitting the software.

VeckoTheGecko:
@surgura do you think this is closed now that #60 is merged? Is there anything pending?

surgura:
yeah I think we are good here

PR #60, opened 2024-08-23
53 interactive loop

surgura:
Sorry this is so large, changes are quite intertwined.

* CLI tool for running expedition (formerly 'cruise')
* Rewrote schedule simulation to be much simpler
* If schedule simulation fails, a checkpoint is created and the user is told to update the schedule and run the tool again, which will resume from the checkpoint.
* Seperated ship configuration and schedule (formerly 'waypoints') into two seperate files with improved loading checks
* One waypoints can have multiple measurements done in parallel
* Fixed some minor bugs that surfaced during creation of new tests

Related to #53


virtual_ship/expedition/simulate_measurements.py
```diff
@@ -0,0 +1,104 @@
...
+    :param measurements: The measurements to simulate.
+    :raises RuntimeError: In case fieldsets of configuration is not provided. Make sure to check this before calling this function.
+    """
+    if isinstance(expedition_dir, str):
+        expedition_dir = Path(expedition_dir)
+
+    if len(measurements.ship_underwater_sts) > 0:
+        print("Simulating onboard salinity and temperature measurements.")
+        if input_data.ship_underwater_st_fieldset is None:
+            raise RuntimeError("No fieldset for ship underwater ST provided.")
```
ammedd, line 39: This comes from the default fieldset? Is that clear with this error? Same for other measurements below

virtual_ship/expedition/verify_schedule.py
```diff
@@ -0,0 +1,130 @@
...
+
+    :param projection: projection used to sail between waypoints.
+    :param ship_config: The cruise ship_configuration.
+    :param schedule: The schedule to verify.
+    :param input_data: Fieldsets that can be used to check for zero UV condition (is waypoint on land).
+    :raises PlanningError: If waypoints are not feasible or incorrect.
+    :raises ValueError: If there are no fieldsets in the ship_config, which are needed to verify all waypoints are on water.
+    """
+    if len(schedule.waypoints) == 0:
+        raise PlanningError("At least one waypoint must be provided.")
```
ammedd, line 32: Are they called waypoints for the users? Else rename here

virtual_ship/expedition/verify_schedule.py
```diff
@@ -0,0 +1,130 @@
...
+            input_data.ctd_fieldset,
+            input_data.drifter_fieldset,
+            input_data.ship_underwater_st_fieldset,
+        ]
+        if fs is not None
+    ]
+    # check if there are any fieldsets, else its an error
+    if len(available_fieldsets) == 0:
+        raise ValueError(
+            "No fieldsets provided to check if waypoints are on land. Assuming no provided fieldsets is an error."
```
ammedd, line 68: Vague error for users

ammedd (commented):
Lot's of changes. Functionality looks good in general, but some things are unclear to me

scripts/download_data.py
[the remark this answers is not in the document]
surgura: I'll change the name to input_data. The script is not really done though, it's just to have at least something there.

tests/expedition/expedition_dir/input_data/.gitignore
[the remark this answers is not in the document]
surgura: download data is not incorporated in the current workflow of the library. it's just a tool that downloads data of the right format to a directory. it is currently up to the user to put it in the right directory. the download_data script is not in a state yet that it makes sense to write tests for it. we should improve the script in a seperate pr

tests/expedition/expedition_dir/schedule.yaml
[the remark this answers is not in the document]
surgura: I tried looking into the argo float, and it seems there is something not working properly in the code. can we postpone this to a next issue? this bug was already there before this pr

tests/expedition/expedition_dir/ship_config.yaml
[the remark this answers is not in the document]
surgura: done

virtual_ship/expedition/checkpoint.py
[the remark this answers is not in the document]
surgura: this is quite hard, as there currently is no way for the schedule simulation to fail. we can add the test once we add some randomness and such

virtual_ship/expedition/do_expedition.py
```diff
@@ -0,0 +1,149 @@
...
+def _get_ship_config(expedition_dir: Path) -> ShipConfig | None:
+    file_path = expedition_dir.joinpath("ship_config.yaml")
+    try:
+        return ShipConfig.from_yaml(file_path)
+    except FileNotFoundError:
+        print(f'Schedule not found. Save it to "{file_path}".')
+        return None
+
+
+def _load_input_data(expedition_dir: Path, ship_config: ShipConfig) -> InputData:
```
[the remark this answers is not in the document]
surgura: I don't understand your question? It checks if config exists for every isntrument, and then tells InputData whether to load the fieldset for instruments that have config

meta.yaml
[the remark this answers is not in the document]
surgura: I made a small fix. does it work now?

virtual_ship/expedition/simulate_measurements.py
surgura: I do not understand your question. The fieldsets must be available in the InputData object. If they are not loaded, this code cannot simulate

virtual_ship/expedition/verify_schedule.py
surgura: they are called waypoints for the users

virtual_ship/expedition/verify_schedule.py
surgura: This will only happen if the user has zero instruments and no fieldsets are loaded

virtual_ship/expedition/checkpoint.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+_YamlDumper.add_representer(
+    InstrumentType, lambda dumper, data: dumper.represent_data(data.value)
+)
+
+
+class Checkpoint(BaseModel):
+    """
+    Checkpoint is schedule simulation.
```
erikvansebille, line 25: I don't understand this comment; expand a bit what is meant?

virtual_ship/expedition/do_expedition.py
```diff
@@ -0,0 +1,149 @@
...
+
+    # calculate expedition cost in US$
+    assert (
+        schedule.waypoints[0].time is not None
+    ), "First waypoint has no time. This should not be possible as it should have been verified before."
+    time_past = schedule_results.time - schedule.waypoints[0].time
+    cost = expedition_cost(schedule_results, time_past)
+    with open(expedition_dir.joinpath("results", "cost.txt"), "w") as file:
+        file.writelines(f"cost: {cost} US$")
+    print(f"This expedition took {time_past} and would have cost {cost:,.0f} US$.")
```
erikvansebille, line 95: Would it not make more sense to display this _after_ the simulate_measurements? Or at least also print it then?

virtual_ship/expedition/input_data.py
```diff
@@ -0,0 +1,173 @@
...
+        # add bathymetry data
+        bathymetry_file = directory.joinpath("bathymetry.nc")
+        bathymetry_variables = ("bathymetry", "deptho")
+        bathymetry_dimensions = {"lon": "longitude", "lat": "latitude"}
+        bathymetry_field = Field.from_netcdf(
+            bathymetry_file, bathymetry_variables, bathymetry_dimensions
+        )
+        # make depth negative
+        if max(bathymetry_field.data > 0):
+            bathymetry_field.data = -bathymetry_field.data
```
erikvansebille, line 112: Could it be that users provide bathymetry that has both positive (for land mountains) and negative (for ocean throughs) values? In that case, this test is not very useful. Perhaps better to check if `min(bathymetry_field.data >= 0)`?

virtual_ship/expedition/ship_config.py
```diff
@@ -0,0 +1,156 @@
+"""ShipConfig and supporting classes."""
+
+from __future__ import annotations
+
+from datetime import timedelta
+from pathlib import Path
+
+import yaml
+from pydantic import BaseModel, ConfigDict, Field, field_serializer
```
erikvansebille, line 9: Is it confusing that Field here is not the same Field object as Parcels uses? that caught me below... Why not import `pydantic` and then refer to `pydantic.Field(...)` etc below?

virtual_ship/expedition/ship_config.py
```diff
@@ -0,0 +1,156 @@
...
+    def from_yaml(cls, file_path: str | Path) -> ShipConfig:
+        """
+        Load config from yaml file.
+
+        :param file_path: Path to the file to load from.
+        :returns: The config.
+        """
+        with open(file_path, "r") as file:
+            data = yaml.safe_load(file)
+        return ShipConfig(**data)
```
erikvansebille, line 156: These were also defined in expedition/schedule.py? Unnecessary replicate here?

virtual_ship/expedition/simulate_measurements.py
```diff
@@ -0,0 +1,104 @@
...
+def simulate_measurements(
+    expedition_dir: str | Path,
+    ship_config: ShipConfig,
+    input_data: InputData,
+    measurements: MeasurementsToSimulate,
+) -> None:
+    """
+    Simulate measurements using parcels.
+
+    Saves everything the $expedition_dir/results.
```
erikvansebille, line 25:
```suggestion
    Saves everything in expedition_dir/results.
```

virtual_ship/expedition/simulate_measurements.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def simulate_measurements(
+    expedition_dir: str | Path,
+    ship_config: ShipConfig,
+    input_data: InputData,
+    measurements: MeasurementsToSimulate,
+) -> None:
+    """
+    Simulate measurements using parcels.
```
erikvansebille, line 23:
```suggestion
    Simulate measurements using Parcels.
```
(we started using capitalised Parcels again...)

virtual_ship/expedition/verify_schedule.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    Check if waypoint is on land by assuming zero velocity means land.
+
+    :param fieldset: The fieldset to sample the velocity from.
+    :param waypoint: The waypoint to check.
+    :returns: If the waypoint is on land.
+    """
+    return fieldset.UV.eval(
+        0,
+        -fieldset.U.depth[0],
```
erikvansebille, line 128: The minus sign is a bit dangerous here; if the grid has not flipped. Perhaps add an `abs()` to force it to always be negative?

tests/expedition/expedition_dir/schedule.yaml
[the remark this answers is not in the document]
surgura: Because the argo float code intentionally deletes argo particles when they go outside of the fieldset

virtual_ship/expedition/checkpoint.py
surgura: Was a typo. is this better?

virtual_ship/expedition/input_data.py
surgura: Actually, I think both checks are very likely to introduce bugs that are hard to debug for someone later. I removed the check altogether and always invert. This is valid because we only support one specific dataset at this point which needs to be inverted. When this class is made more generic, inverting can be an option.

virtual_ship/expedition/ship_config.py
surgura: done

virtual_ship/expedition/ship_config.py
surgura: I think this code is so trivial it does not deserve its own generic implementation. This would made it harder to maintain instead of easier, in my opionion.

virtual_ship/expedition/verify_schedule.py
surgura: improved sampling with new method. but to me it is strange that while we flip the depth, this way of obtaining the depth remains unflipped?

virtual_ship/expedition/input_data.py
```diff
@@ -0,0 +1,176 @@
...
+            "lon": "longitude",
+            "lat": "latitude",
+            "time": "time",
+            "depth": "depth",
+        }
+
+        fieldset = FieldSet.from_netcdf(
+            filenames, variables, dimensions, allow_time_extrapolation=False
+        )
+        fieldset.T.interp_method = "linear_invdist_land_tracer"
```
erikvansebille, line 166: Should fieldset.S not also have the same interp_method?
```suggestion
        fieldset.T.interp_method = "linear_invdist_land_tracer"
        fieldset.S.interp_method = "linear_invdist_land_tracer"
```

VeckoTheGecko (approved):

PR #72, opened 2024-10-21
Add `init` command

### LINKED ISSUE #67: CLI: Generate template config
`virtualship init`: create template yaml file and expedition directory

VeckoTheGecko:
Fixes #67

Changes:
- Add `virtualship init` command
- link to oceanparcels.org in docs sidebar
- add README badges and link to repo
- Fixes:
  - fix documentation config
  - remove `meta.yml` (managed by the feedstock)

---
## Good to know

`virtualship` now comes with a `virtualship.static` folder which is used to store static files (e.g., example configuration `.yml` files). For data files (which are larger, and aren't as core to the functionality of virtualship) shouldn't be put here but instead should be downloaded on demand.

`virtualship.utils` contains a bunch of utilities for `virtualship`, including convenience functions for reading these static files


src/virtualship/cli/commands.py
```diff
@@ -15,7 +17,26 @@
...
-    raise NotImplementedError("Not implemented yet.")
+    path = Path(path)
+    path.mkdir(exist_ok=True)
+
+    config = path / SHIP_CONFIG
+    schedule = path / SCHEDULE
+
+    if config.exists():
+        raise FileExistsError(
+            f"File '{config}' already exist. Please remove it or choose another directory."
```
erikvansebille, line 28: "directory" or "path"?

VeckoTheGecko: Directory. The `path` varname is the path to the directory

erikvansebille (approved):

VeckoTheGecko:
Static files aren't being packaged properly despite working locally. Removing the `--no-build-isolation` flag would fix it, but I'm not 100% sure if that's the right fix. Waiting on reponse in https://github.com/pypa/setuptools/discussions/4694

VeckoTheGecko:
Decided to just remove the flag `--no-build-isolation`

PR #74, opened 2024-10-31
Adding zenodo badge with doi to README

erikvansebille:


ammedd (approved):
Bit sad to see the tests on the newer version failing, but your change has to do with v0.0.2 which should be fine.

PR #75, opened 2024-10-31
Using private grid._depth for negating depth

erikvansebille:
As Parcels [v3.1.0](https://github.com/OceanParcels/Parcels/releases/v3.1.0) has moved from `grid.depth` to `grid._depth`, the code for VirtualShip is now broken. 

This PR fixes that, although a more proper implementation would be to add a `grid.negate_depth()` method to Parcels itself?

### DIFF
--- a/src/virtualship/expedition/input_data.py
+++ b/src/virtualship/expedition/input_data.py
@@ -97,7 +97,9 @@ class InputData:
 
         # make depth negative
         for g in fieldset.gridset.grids:
-            g._depth = -g._depth  # TODO maybe add a grid.negate_depth() method in Parcels?
+            g._depth = (
+                -g._depth
+            )  # TODO maybe add a grid.negate_depth() method in Parcels?
 
         # add bathymetry data
         bathymetry_file = directory.joinpath("bathymetry.nc")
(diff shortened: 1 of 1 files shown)


erikvansebille:
The other issue that causes the CI to fail is not easily fixed within virtualship; it has to do with a bug in a new warning in Parcels, see https://github.com/OceanParcels/Parcels/pull/1747. 

In any case, because it's 'only a warning', it doesn't affect the functionality of virtualship. We probably have to wait until Parcels is patched, @VeckoTheGecko?

ammedd (approved):

VeckoTheGecko:
@erikvansebille I wonder if a better solution would be to pin parcels<3.1.0 for the timebeing? That way we can implement `.negate_depth` and fix the warnings in parcels without having to push out a new release, and virtualship can be good to go for students.

erikvansebille:
Yes I agree; pinning to <3.1.0 (or even 'not 3.1.0'; is that possible?) might be a better solution. Can you implement?

ammedd:
Just so you know. There is no rush to fix things for virtual ship. I plan to have the next group of students working with the code early February.

PR #76, opened 2024-11-05
Add XBT instrument simulation based on CTD

iuryt:
#### Summary
This PR introduces a new instrument simulation, `xbt.py`, located in the `instruments/` directory. The `xbt.py` file is modeled on `ctd.py` and is designed to simulate data collection for Expendable Bathythermographs (XBTs), which capture water temperature at varying depths.

#### Changes
- **New Instrument**: Added `xbt.py` to `instruments/` directory for XBT data simulation.
- **Test**: Added `test_xbt.py` based on `test_ctd.py`
- **Init Update**: Modified `instruments/__init__.py` to include `xbt.py` in the module exports.

#### Notes
- **Status**: Ready to review.
  
#### Next Steps
- Update documentation to reflect the addition of the XBT instrument.

### DIFF
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -46,12 +46,14 @@ def _xbt_cast(particle, fieldset, time):
     )
 
     # delete particle if depth is exactly max_depth
-    if particle.depth + particle_ddepth == particle.max_depth:
+    if particle.depth == particle.max_depth:
         particle.delete()
 
     # set particle depth to max depth if it's too deep
     if particle.depth + particle_ddepth < particle.max_depth:
-        particle_ddepth = particle.max_depth - particle.depth
+        particle_ddepth = particle.max_depth - particle.depth 
+
+
 
 
 def simulate_xbt(
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -93,7 +93,7 @@ def simulate_xbt(
             f"XBT max_depth or bathymetry shallower than maximum {-DT * WINCH_SPEED}"
         )
 
-    # define parcel particles
+    # define xbt particles
     xbt_particleset = ParticleSet(
         fieldset=fieldset,
         pclass=_XBTParticle,
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -51,9 +51,7 @@ def _xbt_cast(particle, fieldset, time):
 
     # set particle depth to max depth if it's too deep
     if particle.depth + particle_ddepth < particle.max_depth:
-        particle_ddepth = particle.max_depth - particle.depth 
-
-
+        particle_ddepth = particle.max_depth - particle.depth
 
 
 def simulate_xbt(
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -97,10 +97,11 @@ def simulate_xbt(
 
     # XBT depth can not be too shallow, because kernel would break.
     # This shallow is not useful anyway, no need to support.
-    if not all([max_depth <= -DT * fall_speed for max_depth, fall_speed in zip(max_depths, initial_fall_speeds)]):
-        raise ValueError(
-            f"XBT max_depth or bathymetry shallower than maximum {-DT * fall_speed}"
-        )
+    for max_depth, fall_speed in zip(max_depths, initial_fall_speeds): 
+        if not max_depth <= -DT * fall_speed
+            raise ValueError(
+                f"XBT max_depth or bathymetry shallower than maximum {-DT * fall_speed}"
+            )
 
     # define xbt particles
     xbt_particleset = ParticleSet(
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -99,7 +99,7 @@ def simulate_xbt(
 
     # XBT depth can not be too shallow, because kernel would break.
     # This shallow is not useful anyway, no need to support.
-    for max_depth, fall_speed in zip(max_depths, initial_fall_speeds): 
+    for max_depth, fall_speed in zip(max_depths, initial_fall_speeds):
         if not max_depth <= -DT * fall_speed
             raise ValueError(
                 f"XBT max_depth or bathymetry shallower than maximum {-DT * fall_speed}"
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -40,8 +40,9 @@ def _xbt_cast(particle, fieldset, time):
     particle_ddepth = -particle.fall_speed * particle.dt
 
     # update the fall speed from the quadractic fall-rate equation
+    # check https://doi.org/10.5194/os-7-231-2011
     particle.fall_speed = (
-        particle.fall_speed - particle.deceleration_coefficient * particle.dt
+        particle.fall_speed - 2 * particle.deceleration_coefficient * particle.dt
     )
 
     # delete particle when it reaches the maximum depth
--- a/src/virtualship/instruments/xbt.py
+++ b/src/virtualship/instruments/xbt.py
@@ -99,8 +99,8 @@ def simulate_xbt(
 
     # XBT depth can not be too shallow, because kernel would break.
     # This shallow is not useful anyway, no need to support.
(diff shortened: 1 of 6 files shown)


iuryt:
It passed the tests (including the one I created for XBT), but I have not tried or implemented it in a tutorial.
This PR is ready to review.

I hope I made your day with this PR, @ammedd ! haha

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,125 @@
...
+)
+
+
+def _sample_temperature(particle, fieldset, time):
+    particle.temperature = fieldset.T[time, particle.depth, particle.lat, particle.lon]
+
+
+def _xbt_cast(particle, fieldset, time):
+    # lowering
+    particle_ddepth = -particle.winch_speed * particle.dt
```
erikvansebille, line 38: Is 'winch' the best word here? An XBT doesn't really have a winch (at least not one on the ship). Use `spooling_speed` or similar?

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,125 @@
...
+    if particle.depth + particle_ddepth < particle.max_depth:
+        particle.raising = 1
```
erikvansebille, line 40: Why set the raising flag here? An XBT doesn't get raised? Should the particle not simply be deleted here?

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    Use Parcels to simulate a set of XBTs in a fieldset.
+
+    :param fieldset: The fieldset to simulate the XBTs in.
+    :param out_path: The path to write the results to.
+    :param xbts: A list of XBTs to simulate.
+    :param outputdt: Interval which dictates the update frequency of file output during simulation
+    :raises ValueError: Whenever provided XBTs, fieldset, are not compatible with this function.
+    """
+    WINCH_SPEED = 1.0  # sink and rise speed in m/s
```
erikvansebille, line 59: Is this the typical fall speed of an XBT? Does it not depend on density? Can we do something more realistic here?

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,125 @@
...
+    ]
+
+    # XBT depth can not be too shallow, because kernel would break.
+    # This shallow is not useful anyway, no need to support.
+    if not all([max_depth <= -DT * WINCH_SPEED for max_depth in max_depths]):
+        raise ValueError(
+            f"XBT max_depth or bathymetry shallower than maximum {-DT * WINCH_SPEED}"
+        )
+
+    # define parcel particles
```
erikvansebille, line 96:
```suggestion
    # define xbt particles
```

erikvansebille (commented):
Nice addition, @iuryt! Thanks for the contribution. Can you lok at these review comments I provided?

src/virtualship/instruments/xbt.py
iuryt: Since XBTs measure temperature and time, what people used to do is to assume a free falling in the form of $z(t) = \alpha t + \beta t^2$.

$\alpha$ depend on the instrument specifications and $\beta$ is usually very small, taking into account the loss of weight. Following Kizu, Yoritaka and Hanawa (2005), $z(t) = 6.828 t - 0.00182 t^2$ for the T-5 (which is the one that goes deepest, as deep as 1830 m) and following Kizu, Sukigara and Hanawa (2011), $z(t) = 6.553 t - 0.00221 t^2$ for the most commonly used T-7 XBT (reaches 760 m).
Since $\beta$ is so small, maybe we can assume that they basically fall at a constant speed $\alpha$?  
What do you think?

Refs:
https://www.aoml.noaa.gov/phod/goos/meetings/2008/XBT/2005JO61115.pdf
https://os.copernicus.org/articles/7/231/2011/

erikvansebille: OK, thanks for clarifying. These are useful papers for background! 

I would stick as close to the 'real deal' as we can, so use a similar nomenclature as in the papers. 

And I agree that the quadratic term is small, but it's simple to calculate (we can keep track of time in the kernel) so why not add the term? Makes the VirtualShip even more realistic ;-)

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,137 @@
...
+    deceleration_coefficient: float
+
+
+_XBTParticle = JITParticle.add_variables(
+    [
+        Variable("temperature", dtype=np.float32, initial=np.nan),
+        Variable("max_depth", dtype=np.float32),
+        Variable("min_depth", dtype=np.float32),
+        Variable("fall_speed", dtype=np.float32),
+        Variable("initial_fall_speed", dtype=np.float32),
```
erikvansebille, line 30: Why does this need to be a separate variable? Can't we just set `fall_speed` above to the initial version at the beginning? Why store it as a separate `Variable`?

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,137 @@
...
+            xbt.max_depth,
+            fieldset.bathymetry.eval(
+                z=0, y=xbt.spacetime.location.lat, x=xbt.spacetime.location.lon, time=0
+            ),
+        )
+        for xbt in xbts
+    ]
+
+    # initial fall speeds
+    initial_fall_speeds = [xbt.initial_fall_speed for xbt in xbts]
```
erikvansebille, line 98:
```suggestion
    initial_fall_speeds = [xbt.fall_speed for xbt in xbts]
```

src/virtualship/instruments/xbt.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # update the fall speed from the quadractic fall-rate equation
+    # check https://doi.org/10.5194/os-7-231-2011
+    particle.fall_speed = (
+        particle.fall_speed - 2 * particle.deceleration_coefficient * particle.dt
+    )
+
+    # delete particle when it reaches the maximum depth
+    if particle.depth + particle_ddepth < particle.max_depth:
+        particle.delete()
```
iuryt, line 50: @erikvansebille 
Now, I am updating the fall speed with the `deceleration_coefficient` and deleting the particle if it reaches the maximum depth. We can also add the papers and the description to the documentation.
I follow the example from `ctd.py`, but I don't see `particle.depth` being updated. Is this correct? How does that work?

erikvansebille: > @erikvansebille Now, I am updating the fall speed with the `deceleration_coefficient` and deleting the particle if it reaches the maximum depth. 

OK

>We can also add the papers and the description to the documentation. 

Yes, good idea to add the references!

> I follow the example from `ctd.py`, but I don't see `particle.depth` being updated. Is this correct? How does that work?

The updating happens through the `particle_ddepth` variable, see the [Kernel loop tutorial](https://docs.oceanparcels.org/en/latest/examples/tutorial_kernelloop.html)

iuryt: Should I "particle.depth = particle_ddepth" ?

erikvansebille: > Should I particle.depth = particle_ddepth ?

No, that's done automatically in the backend; you shouldn't change `particle.depth` at all.

iuryt:
It seems like this is failing the test because it is reaching almost at the maximum depth, but not really there (e.g. `obs_value=7.965900421142578 exp_value=8`).

Do you have any idea what it could be?


```python
FAILED tests/instruments/test_xbt.py::test_simulate_xbts - AssertionError: Observation incorrect xbt_i=0 loc='maxdepth' var='temperature' obs_value=7.965900421142578 exp_value=8.
```
https://github.com/OceanParcels/virtualship/actions/runs/11781564925/job/32816812061#step:5:1886

iuryt:
It was never reaching the first condition, I changed it to particle.depth == particle.max_depth

I ran the pytests and it worked locally. What do you think?

erikvansebille:
Yep that sounds correct.

erikvansebille (approved):

erikvansebille:
CI breaking is because of issue with Parcels v3.1.0; will be fixed by https://github.com/OceanParcels/virtualship/pull/82. 

Shall I merge this PR, @iuryt?

iuryt:
I think we can merge. Based on what @ammedd is doing, I can work on examples and documentation later.
Thanks! I am very happy to collaborate with this project!

PR #80, opened 2024-11-07
added adcp and ctd examples

ammedd:
These work with outdated data from the Virtualship (so some example data is downloaded) and will have to be revised when the tool functions again, but I wanted to get something up for the next classes anyhow.


erikvansebille:
Maybe also remove the 'awesome tutorial' (at https://virtualship--80.org.readthedocs.build/en/80/tutorials/my_tutorial.html)?

erikvansebille (approved):

Issue #68, opened 2024-10-01
CLI: Get data


VeckoTheGecko:
`virtualship fetch`: Downloads the data needed for the simulation as an explicit step so that users can do it separately. If `virtualship` is ran without this, then it will go and fetch this data.

VeckoTheGecko:
Also see if we can use one environment for everything:

```
Data can be downloaded with the `download_data.py` script. For now a different conda env is needed for downloading, see comments in the script.
```

iuryt:
Is someone working on this? If not, I can try to help.
Should we add new variables to schedule, such as `bbox` for the `area_of_interest`?
The rest would be to basically adapt `scripts/download_data.py` to here https://github.com/OceanParcels/virtualship/blob/6f843e969f51cbe6cce0aa374889d3a9b0e49693/src/virtualship/cli/commands.py#L48-L50

Let me know your thoughts.

iuryt:
Yes, I lean more towards adding this to the config file. For example, if we have an experiment that relies only on drifters, the study area and time range will probably be considerably larger than the schedule.

If you haven't started working on this, I can initiate a PR and we can collaborate from there. What do you think?

VeckoTheGecko:
@iuryt Sounds good! Happy for you to do a PR and I can jump in in review

ammedd:
The download_data is all I used before. It was based on an area of interest based on the sampling stations, but indeed, it would be nice to extend this a bit. And in case of Argo/drifter deployment It downloaded a separate dataset that had an area 3degrees more on each side to allow for 3 additional weeks of data from the deployed instruments.

ammedd:
Closed with #83

PR #83, opened 2024-11-13
Add `fetch` command to download expedition data based on `space_time_region` from `schedule.yaml`

iuryt:
Based on the discussion from #68 

### Description
This PR updates the `fetch` command to download data specified by the expedition's `area_of_interest` . Key changes include:

1. **Integration with Schedule File**: The `fetch` command now uses `_get_schedule` to load the `Schedule` YAML file from the specified `expedition_dir`. The `Schedule` now includes `area_of_interest` with spatial and temporal boundaries, which define the download region and timeframe for data subsets.

2. **Unified Data Download Configuration**: All datasets, including bathymetry, are now defined in a single `download_dict`, making the code cleaner and allowing for easy addition or modification of datasets without changes to the function logic.

4. **Consistent Use of `area_of_interest`**: All dataset downloads utilize `area_of_interest` for spatial (`min/max latitude and longitude`) and temporal (`start/end time`) boundaries, ensuring consistency across all downloaded data.

Should we delete the whole `scripts/` folder?

### Status
This is a draft PR. I have not tested running the `fetch` command with an example `expedition_dir` to ensure data is downloaded correctly based on `area_of_interest`.


iuryt:
@ammedd and @VeckoTheGecko
What do you think of this way I am implementing this?
I was also wondering whether _get_schedule and _get_ship_config should be in do_expedition or somewhere else.

src/virtualship/cli/commands.py
```diff
@@ -45,9 +47,76 @@ def init(path):
...
 )
-def fetch(path):
-    """Entrypoint for the tool."""
-    raise NotImplementedError("Not implemented yet.")
+def fetch(expedition_dir: str | Path) -> None:
+    """Entrypoint for the tool to download data based on area of interest."""
+    if isinstance(expedition_dir, str):
+        expedition_dir = Path(expedition_dir)
+
+    # Load schedule
```
VeckoTheGecko, line 55:
```suggestion
```
Code is self documenting here

src/virtualship/cli/commands.py
```diff
@@ -45,9 +47,76 @@ def init(path):
     "path",
     type=click.Path(exists=True, file_okay=False, dir_okay=True, readable=True),
 )
-def fetch(path):
-    """Entrypoint for the tool."""
-    raise NotImplementedError("Not implemented yet.")
+def fetch(expedition_dir: str | Path) -> None:
+    """Entrypoint for the tool to download data based on area of interest."""
+    if isinstance(expedition_dir, str):
+        expedition_dir = Path(expedition_dir)
```
VeckoTheGecko, line 53:
```suggestion
    expedition_dir = Path(expedition_dir)
```
no isinstance check needed. Also `expedition_dir` should be renamed to `path` (so click can function with it)

src/virtualship/cli/commands.py
```diff
@@ -45,9 +47,76 @@ def init(path):
...
+
+    # Extract area_of_interest details from the schedule
+    spatial_range = schedule.area_of_interest.spatial_range
+    time_range = schedule.area_of_interest.time_range
+    start_datetime = datetime.strptime(time_range.start_time, "%Y-%m-%d %H:%M:%S")
+    end_datetime = datetime.strptime(time_range.end_time, "%Y-%m-%d %H:%M:%S")
+
+    # Prompt for user credentials
+    username = input("username: ")
+    password = input("password: ")
```
VeckoTheGecko, line 69: Let's use click to field these inputs

```diff
diff --git a/src/virtualship/cli/commands.py b/src/virtualship/cli/commands.py
index f533b96..800eb0f 100644
--- a/src/virtualship/cli/commands.py
+++ b/src/virtualship/cli/commands.py
@@ -47,7 +47,13 @@ def init(path):
     "path",
     type=click.Path(exists=True, file_okay=False, dir_okay=True, readable=True),
 )
-def fetch(expedition_dir: str | Path) -> None:
+@click.option(
+    "--username", prompt=True, type=str,
+)
+@click.option(
+    "--password", prompt=True, hide_input=True, type=str,
+)
+def fetch(expedition_dir: str | Path, username: str, password: str) -> None:
     """Entrypoint for the tool to download data based on area of interest."""
     if isinstance(expedition_dir, str):
         expedition_dir = Path(expedition_dir)
@@ -64,10 +70,6 @@ def fetch(expedition_dir: str | Path) -> None:
     start_datetime = datetime.strptime(time_range.start_time, "%Y-%m-%d %H:%M:%S")
     end_datetime = datetime.strptime(time_range.end_time, "%Y-%m-%d %H:%M:%S")
 
-    # Prompt for user credentials
-    username = input("username: ")
-    password = input("password: ")
-
     # Define all datasets to download, including bathymetry
     download_dict = {
         "Bathymetry": {

```

VeckoTheGecko (changes requested):
I like the structure. Here's a preliminary review

VeckoTheGecko:
> I was also wondering whether _get_schedule and _get_ship_config

I think `schedule.py` and `ship_config.py` would be more logical places to put them

VeckoTheGecko:
> Should we delete the whole `scripts/` folder?

I think so

---

@iuryt where would the data be downloaded to do you think? I think it would be nice to have it in a subfolder in the mission folder, and also somehow tie in the spatial/temporal ranges to that dataset. Perhaps something like:

```
mission_folder
└── data
    └── vSgsJWk
        ├── a.nc
        ├── b.nc
        ├── c.nc
        ├── d.nc
        └── schedule.yaml
```

where `vSgsJWk` is a hash of the spatial/temporal information. If users change the region/time range it would generate a new folder. Schedule.yaml is just a copy of the file so that users can know what spatial/temporal settings they had.

Not perfect (i.e., changing any of the domains results in a full data download) but at least it avoids re-downloading already downloaded data, and isolates the data files.

src/virtualship/cli/commands.py
```diff
@@ -1,9 +1,11 @@
+from datetime import datetime
 from pathlib import Path
 
 import click
+import copernicusmarine
```
erikvansebille, line 5: Does this then become a dependency in the pyproject.toml and the conda-forge recipe too?

src/virtualship/cli/commands.py
```diff
@@ -45,9 +47,76 @@ def init(path):
...
+            dataset_id=dataset["dataset_id"],
+            variables=dataset["variables"],
+            minimum_longitude=spatial_range.min_longitude,
+            maximum_longitude=spatial_range.max_longitude,
+            minimum_latitude=spatial_range.min_latitude,
+            maximum_latitude=spatial_range.max_latitude,
+            start_datetime=start_datetime,
+            end_datetime=end_datetime,
+            minimum_depth=0.49402499198913574,
+            maximum_depth=5727.9169921875,
```
erikvansebille, line 108: Do we always need to download full-depth datasets? @ammedd, would there also be cases (when not using CTD?) that only upper/surface ocean data are needed? Or would that just overcomplicate the script?

src/virtualship/expedition/area_of_interest.py
```diff
@@ -0,0 +1,30 @@
+"""AreaOfInterest class."""
```
erikvansebille, line 1: "Area" may be a bit vague term. Do we use the word `Area`, `Region` or even `SpaceTimeRegion` throughout? Which term makes most sense? @ammedd, which would you prefer?

src/virtualship/static/schedule.yaml
```diff
@@ -1,3 +1,12 @@
+area_of_interest:
+  spatial_range:
+    minimum_longitude: -5
+    maximum_longitude: 5
```
erikvansebille, line 4: Do we need to check here that Longitude is either between -180 and 180, or between 0 and 360? Do we need to add a (private) method to the code to make sure this is converted in the way that copernicusmarine expects the longitude range to be? or can that handle both options?

src/virtualship/cli/commands.py
erikvansebille: Great idea! And also support credientials stored in a file like in https://help.marine.copernicus.eu/en/articles/8185007-copernicus-marine-toolbox-credentials-configuration?

erikvansebille: And perhaps also provide a (short) message how to obtain these credentials? Via https://data.marine.copernicus.eu/register?

erikvansebille (commented):
Looks good! More comments from me below

src/virtualship/cli/commands.py
VeckoTheGecko: and as a dependency in `environment.yml`

src/virtualship/cli/commands.py
iuryt: We can add this very easily to the `area_of_interest` . Do we have checks to see if the data covers the survey?

iuryt:
> > Should we delete the whole `scripts/` folder?
> 
> I think so
> 
> @iuryt where would the data be downloaded to do you think? I think it would be nice to have it in a subfolder in the mission folder, and also somehow tie in the spatial/temporal ranges to that dataset. Perhaps something like:
> 
> ```
> mission_folder
> └── data
>     └── vSgsJWk
>         ├── a.nc
>         ├── b.nc
>         ├── c.nc
>         ├── d.nc
>         └── schedule.yaml
> ```
> 
> where `vSgsJWk` is a hash of the spatial/temporal information. If users change the region/time range it would generate a new folder. Schedule.yaml is just a copy of the file so that users can know what spatial/temporal settings they had.
> 
> Not perfect (i.e., changing any of the domains results in a full data download) but at least it avoids re-downloading already downloaded data, and isolates the data files for different areas.

Thank you so much for your suggestions! I’ve been thinking about how to put them into action.

One idea I had is to keep all the data in a single folder and just check the metadata for any changes in the spatial-temporal domain. If the user changes the domain, we can easily overwrite the data. I’m curious, though—why might someone need access to the previous data? To make things easier and skip loading the metadata each time, we could simply check if the copied `data/schedule.yaml` matches the current `schedule.yaml` domain.

I really like the concept of using a hash, but I’m a bit concerned it might not be the friendliest option for those who aren’t familiar with it, especially students. It might require some extra explanation during lectures. I’d love to hear what @ammedd and @erikvansebille think about this!

erikvansebille:
> I really like the concept of using a hash, but I’m a bit concerned it might not be the friendliest option for those who aren’t familiar with it, especially students. It might require some extra explanation during lectures. I’d love to hear what @ammedd and @erikvansebille think about this!

Hi @iuryt, I agree that a hash-based folder name might not be very friendly to work with for students. Perhaps a timestamp-based folder would be friendlier? E.g. `mission_folder/data/download_YYYMMDD_HHMMSS/*`. 

That would also help easily locate the latest version of the downloaded data because the downloads would automatically be sorted.

erikvansebille:
> But do you think it is helpful to keep past downloads?

Well, we don't want people to have to redownload every time they run the virtual ship, I assume. We could only keep the last download, but then why even have a folder per download (the idea of a hash in the first place)?

iuryt:
> > But do you think it is helpful to keep past downloads?
> 
> Well, we don't want people to have to redownload every time they run the virtual ship, I assume. We could only keep the last download, but then why even have a folder per download (the idea of a hash in the first place)?

I was wondering if we could check for updates to the `schedule.yaml` file compared to the copy in the download folder. We could only overwrite it if there were changes in the spatial-temporal domain. This way, we wouldn’t need separate folders for each download. Let me know your thoughts.

VeckoTheGecko:
I'm still a fan of hashes as I think it simplifies the implementation details and adds a bit of futureproofing (i.e., what if we want to "force" new data downloads with a new version of virtualship because we add a new dataset, or the download command has changed. Doing so with a hash we can change a single line of code in the hashing function. Also what if a new version of VS changes the format of `schedule.yaml`?). The concept of hashing isn't important to students, all they need to know is: "Defining the spatial/temporal domain in `schedule.yaml` will make it so that data is downloaded to a unique folder. If you re-use the same domain previous downloads will be found". If users want to look at the details of the download, they can look into the folder at `schedule.yaml`. Doing a purely timestamp based system would require loading each schedule, which can be sidestepped via hashing. We could have the best of both worlds by doing `data/YYYYMMDD_HHMMSS_vSgsJWk` so its human readable and functional with the hash.

This is all underpinned by whether re-using the data is important. My conversations with @ammedd mentioning that these data downloads as sizeable and take a long time (if someone can advise how long/how large, since I don't know this) leads me to be inclined to save the downloads and re-use them as much as practical, and leave managing the storage space to the user. A "one folder per domain" approach takes more space, but allows the user to change the domain as they see fit without worrying about it being cleared out from a separate run they did.

> One idea I had is to keep all the data in a single folder

As in, across different missions the data would also be saved to the same folder? (e.g., a folder that they're prompted to specify the first time that they run VS?) I initially suggested in https://github.com/OceanParcels/virtualship/pull/83#issuecomment-2478056334 the data being downloaded in the mission folder so that it was very visible to the user and centralised. I think the mission folder approach is easier to implement (no need to worry about (a) permissions elsewhere on the file system, (b) users not knowing where the data is stored, (c) setting/saving application level configuration), although users would need to manually copy data files between mission if they want to avoid doing a download. I think that trade-off would be worth it

src/virtualship/cli/commands.py
ammedd: The are indeed (many) cases where the full-depth is not needed. It would be good to check the ship_config for the maximum depth needed.

src/virtualship/expedition/area_of_interest.py
ammedd: Yes, let's do `SpaceTimeRegion`

ammedd:
Wow! So many ideas/much work going on!

I like saving data to the mission folder and the `data/YYYYMMDD_vSgsJWk` human readable and hash option sounds like the best of both worlds. Stating the date of the downloaded dataset, not the date of the download ;-)

If I remember correctly downloads took up to 2 hours each last year and my 256GB disk couldn't handle all the data needed for the 7 groups we had last year. I do think the download method (and speed) has improved since then.

erikvansebille:
Maybe this is because the PR is still in draft, but when I tried to use it (to download data for a drifter simulation), I got the following error
```bash
(parcels) erik ~/Codes/VirtualShip % virtualship fetch TrAtldrifters
Traceback (most recent call last):
  File "/Users/erik/anaconda3/envs/parcels/bin/virtualship", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: fetch() got an unexpected keyword argument 'path'
```

Any idea what's going on?

src/virtualship/cli/commands.py
```diff
@@ -45,9 +47,74 @@ def init(path):
     "path",
     type=click.Path(exists=True, file_okay=False, dir_okay=True, readable=True),
 )
-def fetch(path):
-    """Entrypoint for the tool."""
-    raise NotImplementedError("Not implemented yet.")
+def fetch(expedition_dir: str | Path) -> None:
+    """Entrypoint for the tool to download data based on area of interest."""
```
VeckoTheGecko, line 51: its here, click expects the parameter name to be `path` to match the decorator

erikvansebille: OK, so am I doing something wrong then when I do `virtualship fetch TrAtldrifters` (and `TrAtldrifters` is the name of a directory that exists)?

VeckoTheGecko: No, its an error in the code

```suggestion
def fetch(path: str | Path) -> None:
    """Entrypoint for the tool to download data based on area of interest."""
```
+ an update to the other references. Should I push a patch for this? (+ my other review comments)

erikvansebille: yes please!

VeckoTheGecko:
> ship_config



> > I was also wondering whether _get_schedule and _get_ship_config
> 
> I think `schedule.py` and `ship_config.py` would be more logical places to put them

On second thought, I think let's defer this decision. I think splitting these similar functions into different files has it's own cost.

src/virtualship/cli/commands.py
VeckoTheGecko: Done in 391f5ef161a303c6e0e98f2d6fc6f0d3bc05950c

src/virtualship/cli/commands.py
VeckoTheGecko: Done in c691a478fea5443d3680152c5c1e0c1e9bc6f14c

src/virtualship/cli/commands.py
VeckoTheGecko: Resolved now

iuryt:
Hi all, sorry for not finishing this. I have a deadline for a paper revision this Friday and will work on this PR right after that.

VeckoTheGecko:
All good. Let me know if there's anything I can do to help :)

erikvansebille:
Note that I just got an email from Copernicus Marine Service that they are about to update their `copernicusmarine` to v2.0.0, which has some breaking changes: https://help.marine.copernicus.eu/en/articles/9978784-what-s-new-in-the-pre-release-of-version-2-0-0-of-the-copernicus-marine-toolbox. 

So it might be good to already implement the _new_ API in `virtualship fetch`

VeckoTheGecko:
Good to know about v2! I didn't see a timeline for the release of v2, only "soon". I think it would be good to pin <2 for the timebeing so that we aren't coupled to their development and add v2 support later? Takes out the uncertainty especially around the holiday season.

iuryt:
The current version is, at least, working, but so far, it is not creating a separate folder for each update on schedule.
I think we can add a 

```python
from getpass import getpass
password = getpass()
```
for the case the person doesn't add the `--password` key.


```bash
(ship) iury@coonamessett:~/projects/active/tmp$ virtualship init test/
Created 'ship_config.yaml' and 'schedule.yaml' at test.
(ship) iury@coonamessett:~/projects/active/tmp$ cd test/
(ship) iury@coonamessett:~/projects/active/tmp/test$ virtualship fetch ./ --username <myusername> --password <mypassword>
WARNING - 2024-12-21T17:23:32Z - 'force_dataset_part' has been deprecated. Use 'dataset_part' instead. This option will no longer be available in copernicusmarine>=2.0.0. Please refer to the documentation when the new major version is released for more information.
Fetching catalog: 100%|█████████████████████████████████████████████| 3/3 [00:22<00:00,  7.52s/it]
INFO - 2024-12-21T17:23:59Z - Dataset version was not specified, the latest one was selected: "202311"
INFO - 2024-12-21T17:23:59Z - You forced selection of dataset part "bathy"
INFO - 2024-12-21T17:23:59Z - Service was not specified, the default one was selected: "static-arco"
INFO - 2024-12-21T17:24:01Z - Downloading using service static-arco...
INFO - 2024-12-21T17:24:02Z - Estimated size of the dataset file is 0.056 MB.
INFO - 2024-12-21T17:24:02Z - Writing to local storage. Please wait...
100%|███████████████████████████████████████████████████████████████| 6/6 [00:00<00:00,  7.07it/s]
INFO - 2024-12-21T17:24:03Z - Successfully downloaded to bathymetry.nc
WARNING - 2024-12-21T17:24:03Z - 'force_dataset_part' has been deprecated. Use 'dataset_part' instead. This option will no longer be available in copernicusmarine>=2.0.0. Please refer to the documentation when the new major version is released for more information.
INFO - 2024-12-21T17:24:05Z - Dataset version was not specified, the latest one was selected: "202406"
INFO - 2024-12-21T17:24:05Z - Dataset part was not specified, the first one was selected: "default"
INFO - 2024-12-21T17:24:07Z - Service was not specified, the default one was selected: "arco-time-series"
INFO - 2024-12-21T17:24:09Z - Downloading using service arco-time-series...
INFO - 2024-12-21T17:24:12Z - Estimated size of the dataset file is 530.876 MB.
INFO - 2024-12-21T17:24:12Z - Writing to local storage. Please wait...
100%|██████████████████████████████████████████████████████| 12314/12314 [00:34<00:00, 353.42it/s]
INFO - 2024-12-21T17:24:58Z - Successfully downloaded to default_uv.nc
WARNING - 2024-12-21T17:24:58Z - 'force_dataset_part' has been deprecated. Use 'dataset_part' instead. This option will no longer be available in copernicusmarine>=2.0.0. Please refer to the documentation when the new major version is released for more information.
INFO - 2024-12-21T17:24:59Z - Dataset version was not specified, the latest one was selected: "202406"
INFO - 2024-12-21T17:24:59Z - Dataset part was not specified, the first one was selected: "default"
INFO - 2024-12-21T17:25:01Z - Service was not specified, the default one was selected: "arco-time-series"
INFO - 2024-12-21T17:25:03Z - Downloading using service arco-time-series...
INFO - 2024-12-21T17:25:06Z - Estimated size of the dataset file is 265.438 MB.
INFO - 2024-12-21T17:25:06Z - Writing to local storage. Please wait...
100%|████████████████████████████████████████████████████████| 6158/6158 [00:15<00:00, 386.82it/s]
INFO - 2024-12-21T17:25:27Z - Successfully downloaded to default_s.nc
WARNING - 2024-12-21T17:25:27Z - 'force_dataset_part' has been deprecated. Use 'dataset_part' instead. This option will no longer be available in copernicusmarine>=2.0.0. Please refer to the documentation when the new major version is released for more information.
INFO - 2024-12-21T17:25:28Z - Dataset version was not specified, the latest one was selected: "202406"
INFO - 2024-12-21T17:25:28Z - Dataset part was not specified, the first one was selected: "default"
INFO - 2024-12-21T17:25:31Z - Service was not specified, the default one was selected: "arco-time-series"
INFO - 2024-12-21T17:25:33Z - Downloading using service arco-time-series...
INFO - 2024-12-21T17:25:35Z - Estimated size of the dataset file is 265.438 MB.
INFO - 2024-12-21T17:25:35Z - Writing to local storage. Please wait...
100%|████████████████████████████████████████████████████████| 6158/6158 [00:16<00:00, 370.47it/s]
INFO - 2024-12-21T17:25:57Z - Successfully downloaded to default_t.nc
Data download based on area of interest completed.
```

iuryt:
The current version is running locally.
It checks if the user gives username and/or password, if not, it will ask for it.
It also automatically creates a `data` folder and its sha256 subfolder based on the area of interest from `schedule.yaml` and then download the data to the subfolder. If the data is already there, it will overwrite it.
This all seems to be working properly.

We need to add the check of the domain bounds based on the waypoints and if the user does not provide minimum and maximum depth, estimate it based on the used instruments.

@ammedd , @VeckoTheGecko  and @erikvansebille 
Let me know what you think and test it running locally to see if this is working as expected.

I might be able to work on the other points on Monday.

Another thing we need to do is to add an unit test for the `fetch` command, which is tricky because it depends on an authentication to work properly. Any ideas?

erikvansebille:
Thanks for this work, @iuryt! I checked the `fetch` script and it works for me. Yay!

Note that we use negative depths throughout in `virtualship` (for pedagogical purposes, because that is more mathematically consistent), so you may also want to require a `maximum_depth` in `schedule.yaml` that's negative. 
Perhaps we should even steer away completely from `mim(mum)_depth` and `max(imum)_depth`, and use e.g. `shallowest_depth` and `deepest_depth` instead, throughout the tool? That avoids confusion when max < min. 

In any case, not that most of the team will be free from tomorrow until 6 January, as Utrecht University closes for the end-of-year period and we all have to take compulsory leave. So don't expect much engagement from @ammedd, @VeckoTheGecko  and myself over the coming days...

iuryt:
I agree with your ideas. Happy holidays!
We can discuss the other points in January.

VeckoTheGecko (changes requested):
Thanks for the work @iuryt! Good to see that its working for @erikvansebille . Just some review feedback here from looking at the code.

We also still need to implement the caching itself (ie., if the item has been downloaded before, then it is skipped). This doesn't look to be inbuilt into copernicusmarine. Then we'd also need to consider how partial downloads are handled (e.g., a user cancels a download midway, the data is incomplete but the folder is created - perhaps we can add a `virtualship_log.yaml` file to the download folder with `status: complete`). Let me know if there are aspects that you want me to take a look at :)

src/virtualship/static/schedule.yaml
VeckoTheGecko: > this is converted in the way that copernicusmarine expects the longitude range to be? or can that handle both options?

Looked through their docs, and they're quite versatile with how longitude is provided. I propose we do [-180,180] for now (missions that cross the international dateline would probably be another issue)

VeckoTheGecko:
I'm going to continue with development on this PR so that we can get it ready for the next batch of students in February. Thanks for all your work so far @iuryt ! Let me know if theres anything specific that you would like to work on here, or what your time commitment looks like for this.

src/virtualship/cli/commands.py
VeckoTheGecko: Is this something that we should deal with in this PR? It would require some choices about depths for which datasets/instruments, which might be better to discuss in another issue/future PR.

VeckoTheGecko:
Also mentioned in https://github.com/OceanParcels/virtualship/pull/83#discussion_r1913431590, but wondering:

Should we leave the "only downloading up to a certain depth" for a future issue and PR so that we can scope out and discuss it separately? If I'm not mistaked, from our discussion earlier, it didn't sound very simple what the answer should be.

erikvansebille (commented):
Do we also need to update the information on docs.oceanparcels.org, via changes to `README.md`? E.g. give more info on the credentials in the `virtualship fetch --help` info

erikvansebille:
I just tried running the `virtual ship` fetch command, using the following `schedule.yaml`
```yaml
area_of_interest:
  spatial_range:
    minimum_longitude: 5
    maximum_longitude: 60
    minimum_latitude: -45
    maximum_latitude: -30
    minimum_depth: -30
    maximum_depth: 0
  time_range:
    start_time: 2023-07-01 00:00:00
    end_time: 2023-09-30 00:00:00
waypoints:
  - instrument: DRIFTER
    location:
      latitude: -42
      longitude: 10
    time: 2023-07-01 01:00:00
```
but got the error below
```bash
(parcels) erik ~/Codes/VirtualShip % virtualship fetch TrAtldrifters
Credentials not provided. Credentials can be obtained from https://data.marine.copernicus.eu/register. Either pass in via `--username` and `--password` arguments, or via config file at 'TrAtldrifters/credentials.yaml'. Config file should be YAML along following format:
### TrAtldrifters/credentials.yaml

COPERNICUS_PASSWORD: my_password
COPERNICUS_USERNAME: my_username

###

Prompting for credentials instead...

username: evansebille1
password: 
Traceback (most recent call last):
  File "/Users/erik/anaconda3/envs/parcels/bin/virtualship", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/erik/Codes/VirtualShip/src/virtualship/cli/commands.py", line 140, in fetch
    copernicusmarine.subset(
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/copernicusmarine/core_functions/deprecated.py", line 136, in wrapper
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/copernicusmarine/python_interface/exception_handler.py", line 17, in wrapper
    raise exception
  File "/Users/erik/anaconda3/envs/parcels/lib/python3.12/site-packages/copernicusmarine/python_interface/exception_handler.py", line 13, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: subset() got an unexpected keyword argument 'overwrite'
```

Any idea what's going wrong?

src/virtualship/cli/_creds.py
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+
+def get_credentials_flow(
+    username: str | None, password: str | None, creds_path: Path
+) -> tuple[str, str]:
+    """
+    Execute flow of getting credentials for use in the `fetch` command.
+
+    - If username and password are provided via CLI, use them (ignore the credentials file if exists).
```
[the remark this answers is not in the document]
VeckoTheGecko: This was intended more as developer documentation rather than user documentation. What do we think in terms of user documentation? Perhaps I can write a documentation page? (I imagine that we wouldn't want this information silo'd in a notebook - depending on the structure of the notebook, but I imagine there would be other course content in there)

Updating the docstring here wont appear to users unless its also displayed in some sort of user documentation.

src/virtualship/cli/commands.py
```diff
@@ -38,23 +48,137 @@ def init(path):
...
+def fetch(path: str | Path, username: str | None, password: str | None) -> None:
+    """
+    Download input data for an expedition.
+
+    Entrypoint for the tool to download data based on space-time region provided in the
+    schedule file. Data is downloaded from Copernicus Marine, credentials for which can be
+    obtained via registration: https://data.marine.copernicus.eu/register . Credentials can
+    be provided on prompt, via command line arguments, or via a YAML config file. Run
+    `virtualship fetch` on a expedition for more info.
+    """
```
VeckoTheGecko, line 77: > Do we also need to update the information on docs.oceanparcels.org, via changes to `README.md`? E.g. give more info on the credentials in the `virtualship fetch --help` info

oh yes, that was an item discussed yesterday. Updated - thoughts on this (which will be surfaced in the --help command)? I don't want to double up on the info, hence the `Run
    `virtualship fetch` on a expedition for more info.`. Not sure if `Credentials can
    be provided on prompt` can be worded better

VeckoTheGecko: I'll update the readme once confirmed

VeckoTheGecko:
By the way, I have also updated all references to "area of interest" to "space time region" as per the review feedback before

erikvansebille (approved):

PR #87, opened 2024-11-19
Add assignment for an ECR expedition proposal

ammedd:
This assignment is designed for graduate students with little former experience in python and should take 10 to 20 hours to complete.


ammedd:
Hi @iuryt. Thanks for thinking along!

Yes, let's move all the case studies to a notebook on git. That way anyone can add points. I'll create an issue for that and we'll see who gets to it first. 

I'll add the XBT option to the text.

erikvansebille (commented):
Notebook looks great! Below some comments (because I can't comment on individual cells in the notebook file directly):

- [x]  Move the install cell to further down (until it is actually needed), so that readers don't get distracted by it?
- [ ]  The "here's a list of interesting regions" should be a link?
- [ ]  Under ADCP, is the concept of a "transect" clear to readers? Perhaps define that first?
- [ ]  The list to calculate time needed for CTD doesn't parse well on readthedocs. And should it be 'maximum depth' (which could be water depth, but also shallower)?
- [ ]  In the drifters section, change "temperature" to "sea surface temperature"?
- [ ]  The Argo floats section mentions tutorial 2, but is there such a numbering of tutorials/ Better to make it a link?
- [ ]  In the paragraph about lead times, can students know what a 'larger expedition' is? Where is the difference with a regular expedition?
- [ ]  In the last paragraph before Outline, also mention that working in ice-covered areas required special ships? And is typically only done in local summer?
- [ ]  In Outline 2; why would student numbers be important here?
- [ ]  In Outline 6, Underway data does not start on a new line
- [ ]  In Outline Cruise Data and Outline 11, the link does not contain the "/cruiselocationplanning#" part
- [ ] In Outline 7; this suggests students can plan cruises into the future (also related to lead times), but we will not have hydrodynamic data into the future. Make clear somewhere that the virtual cruises must happen between 2021(?) and 2024, so we can use CMS data?
- [ ] In outline 8: what is the "working area"? The map of the ocean? Or the layout of the ship decks?
- [ ] Point 12 Feasibility is a subsection instead of a subsubsection?

ammedd:
Thanks. Lot's of valid points to improve upon. 
Added some info from last year on logistics also. 

There are some formatting issues that I'm not sure how to fix. In VScode they look perfect, but on read-the-docs some don't come through, neither in google colab. It's not clear to me what fails where and how to fix it, so I'll need a few tries...

I'll leave the point about Outline 7 open for now, because I want to use the notebook in a tutorial this week and students can (and should?) plan into the future. I need a bit more time to consider how to phrase this in the context of the virtual ship package. I'll add it to the discussion page.

VeckoTheGecko:
Rebasing this turned out to be a lot harder than I initially thought. I've made another PR for this #113 , which I think is less error prone than an improper rebase.

PR #94, opened 2024-11-28
Add Argo tutorial

ammedd:
Adding a tutorial based on Parcels for people to learn about "programming" the Argo floats.


erikvansebille (commented):
Looks good in principle. But why not use the VirtualShip Argo capabilities here? Why keep with the raw Parcels version? Is that intentional?

Below some specific comments
- [ ] Why is the Research_proposal_only.ipynb file also part of this PR? Was that not a different PR?
- [ ] Provide some information on the region of the data at the beginning? Mention that this is in the Tropical Atlantic (just off the coast of Brazil?)
- [ ] The double quotes in the italic Notes before the ParticleClass and execute cells don't parse well. Make them single quotes?
- [ ] Add a markdown cell at the end with some questions to prompt readers to interpret what they see?

ammedd:
Superseded by #132

PR #95, opened 2024-11-29
Using the new grid.negate_depth() method

erikvansebille:
Note that this will only work after Parcels v3.1.1 is released and on conda. But create this PR now already so we don't forget to update the VirtualShip


VeckoTheGecko (approved):
https://github.com/conda-forge/virtualship-feedstock/pull/2

PR #102, opened 2025-01-16
Add contributing section and update README help messages

VeckoTheGecko:
I thought it would be nice to add a contributing section like that seen in [parcels README](https://github.com/OceanParcels/parcels?tab=readme-ov-file#contributors)

- [x] Re-ran README.md help commands if the CLI has changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,3 +114,12 @@ The scripts are written to work with [A-grid ocean data from the Copernicus Mari
 ## Source code
 
 The code for this project is [hosted on GitHub](https://github.com/OceanParcels/virtualship).
+
+### Contributors
+
+<a href="https://github.com/oceanparcels/virtualship/graphs/contributors">
+  <img src="https://contrib.rocks/image?repo=oceanparcels/virtualship" />
+</a>
+
+**All contributions are welcome! See the [contributing page](https://virtualship.oceanparcels.org/en/latest/contributing.html) in our documentation to see how to get involved.**
+Image made with [contrib.rocks](https://contrib.rocks).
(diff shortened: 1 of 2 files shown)


src/virtualship/cli/commands.py
```diff
@@ -174,11 +174,6 @@ def fetch(path: str | Path, username: str | None, password: str | None) -> None:
     "path",
     type=click.Path(exists=True, file_okay=False, dir_okay=True, readable=True),
 )
-@click.option(
-    "--username",
-    prompt=True,
-    type=str,
-)
```
VeckoTheGecko, line 181: This snuck in with the last PR, its redundant

README.md
```diff
@@ -114,3 +122,12 @@ The scripts are written to work with [A-grid ocean data from the Copernicus Mari
 ## Source code
 
 The code for this project is [hosted on GitHub](https://github.com/OceanParcels/virtualship).
```
erikvansebille, line 124: Can we also have a GitHub logo on the top-right, like in https://docs.oceanparcels.org/en/latest/index.html?

erikvansebille (approved):

README.md
VeckoTheGecko: This uses a different theme which, unfortunately, doesn't allow the customisation of these buttons. Perhaps we can add a link ("Project on GitHub") to the left sidebar? There is already a github link in the footer, though that is admittedly out of the way.

erikvansebille: OK clear. But why don’t we use the same theme as Parcels itself? Is there a good reason for that? One theme would be nice for consistency/recognition, and would also reduce maintenance (a bit)?

VeckoTheGecko: > Is there a good reason for that?

Nope

> One theme would be nice for consistency/recognition, and would also reduce maintenance (a bit)?

Yes, sounds good. I can work on a migration

PR #111, opened 2025-01-28
Converting MFP CSV to YAML schedule

### LINKED ISSUE #61: YAML schedule from MFP CSV
So the tool parses csv output from https://marinefacilitiesplanning.com/userportal

iuryt:
@ammedd @VeckoTheGecko 

The file does not contain the time expected for each of the stations, how should we guess this?

Also, how should we organize this code? Should we embed this to the `virtualship init` command?
e.g. `virtualship init --mfp_file ./CoordinatesExport-Filled.xlsx`

So far, this is just a script that creates the `yaml` file. We can delete them after deciding where to implement it.

I am also using the CSV version to avoid installing `openpyxl`, but we can also install it, no problem.

- [x] Closes #61 
- [x] Tests added
- [ ] Re-ran README.md help commands if the CLI has changed


ammedd:
I'll ask the students to add a column with station time. I hope it will become available in later downloads from the MFP website directly as well. It is included online. 

`virtualship init --mfp_file ./CoordinatesExport-Filled.xlsx` sounds great! maybe changing `file` for csv or DD
When you download from MFP, you choose Export Coordinates - DD (decimal degrees). There are other options but I would not cater for those at the moment.

iuryt:
@ammedd 
Can you give me a version of this file with the times using the template you plan to use?

VeckoTheGecko:
Okay, I've managed to look through the code. Thanks for picking this up @iuryt ! I definitely see how this will streamline things for users - especially for the waypoints and the bounding box.

> But that means that we need to either leave blank for the dates or guess it, right? There are also stations that will have more than one instrument, but it will be hard to guess they are the same if we leave the dates blank. Of course, they have the same location, but if longitudes and latitudes are too close, it might be harder than just adding a column to the CSV. Although I understand that adding a column manually is more prone to error.

Correct, we'll have to leave the dates blank - its just information that is not provided from the MFP export. Though the order of points in the YAML should match the CSV. If the MFP export changes in future, then we can adapt. So yes, it would also be difficult to guess whether stations are the same. Honestly, I think that we should avoid making that guess. Making it clear that `virtualship init --from-mfp` will only populate some of the fields, and the user will be required to take it from there (e.g., merging waypoints to the same location that were initially spaced a bit further apart in MFP, filling in the dates from the UI of MFP - since they weren't included in the export).

Its not optimal, but given MFP->virtualship isn't necessarily 1to1, and given the limited export, I think that its all we can do. I also think this is the clearer approach from a maintenance POV - being explicit with what is supported, and by not making additional assumptions which may not be right - and hands the control to the user to modify the schedule file.

> Also, how should we organize this code? Should we embed this to the `virtualship init` command?
> e.g. `virtualship init --mfp_file ./CoordinatesExport-Filled.xlsx`

Sounds like a good plan! My vote is for `--from-mfp` since that is more in line with CLI conventions. We can then add in the help message something like this would be good:

```
--from-mfp    Partially initialise a project from an exported xlsx or csv file from NIOZ' Marine Facilities Planning tool (specifically the "Export Coordinates > DD" option). User edits are required after initialisation.
```

> I am also using the CSV version to avoid installing `openpyxl`, but we can also install it, no problem.


Let's install `openpyxl`

---
Other notes:

- Could we add an error message to users if the CSV headers are not as expected?
  - Incorrect headers -> `Error: "Found headers [...], but expected headers [...]. Are you sure that you're using the correct export from MFP?"`
  - Additional headers -> `Warning: "Found additional unexpected headers [...]. Manually added columns have no effect. If the MFP export format changed, please submit an issue https://github.com/OceanParcels/virtualship/issues."`
- It would be nice to have a message at the end that its up to the user to populate the rest of the YAML
- Would you be comfortable writing some unit tests for this functionality as well?

iuryt:
@ammedd and @VeckoTheGecko 

But I tested it, and the current version is working.
I also added some `maximum_depth` and `buffer` for longitude and latitude that depend on the instruments.

```python
    # Define maximum depth and buffer for each instrument
    instrument_properties = {
        "CTD": {"depth": 5000, "buffer": 1},
        "DRIFTER": {"depth": 1, "buffer": 5},
        "ARGO_FLOAT": {"depth": 2000, "buffer": 5},
    }
```

Let me know what you think.

I have not added yet
 - [ ] Error message to users if the CSV headers are not as expected
 - [ ] Additional headers -> Warning
 - [ ] Message at the end that its up to the user to populate the rest of the YAML
 - [ ] Unit tests for this functionality

erikvansebille:
Looks goo, but perhaps explain what `buffer` means in the dictionary above?

src/virtualship/utils.py
VeckoTheGecko, line 1: Do you think it would be possible to use the Pydantic objects directly here? (Waypoint, SpaceTimeRegion, then building up to Schedule) I think it would be less error prone than manually creating the dictionary/YAML.

src/virtualship/utils.py
```diff
@@ -37,3 +40,99 @@ def _dump_yaml(model: BaseModel, stream: TextIO) -> str | None:
...
+    instrument_properties = {
+        "CTD": {"depth": 5000, "buffer": 1},
+        "DRIFTER": {"depth": 1, "buffer": 5},
+        "ARGO_FLOAT": {"depth": 2000, "buffer": 5},
+    }
+
+    # Extract unique instruments from dataset
+    unique_instruments = np.unique(
+        np.hstack(coordinates_data["Instrument"].apply(lambda a: a.split(", ")).values)
+    )
```
VeckoTheGecko, line 78: I think this would just be clearer as a for loop.

```py
unique_instruments = set()
for row in ....:
    instruments = ...split(", ")
    unique_instruments |= set(instruments)  # or just `unique_instruments = unique_instruments | set(instruments)`
```
That way there's no need to write the comment since its evident from the code

src/virtualship/cli/commands.py
```diff
@@ -43,7 +53,16 @@ def init(path):
...
 
     config.write_text(utils.get_example_config())
-    schedule.write_text(utils.get_example_schedule())
+
+    if mfp_file:
+        # Generate schedule.yaml from the MPF file
+        click.echo(f"Generating schedule from {mfp_file}...")
+        mfp_to_yaml(
+            mfp_file, str(path)
+        )  # Pass the path to save in the correct directory
```
VeckoTheGecko, line 62: If `mfp_to_yaml` returned a string representing the yaml instead of doing the file modifications, then we can use `schedule.write_text(...)` like in the else clause. So something like

```py
if mfp_file:
    # don't repeat the comment here :) the log message is comment enough
    click.echo(f"Generating schedule from {mfp_file}...")
    schedule_body = mfp_to_yaml(
        mfp_file,
    )
else:
    schedule_body = utils.get_example_schedule()
schedule.write_text(...)

```

environment.yml
```diff
@@ -12,6 +12,7 @@ dependencies:
   - pip
   - pyyaml
   - copernicusmarine >= 2
+  - openpyxl >= 3.1.5
```
VeckoTheGecko, line 15:
```suggestion
  - openpyxl
```
I don't think we need to specify the lower bound. Its best to be flexible unless there is an incompatibility with x version in our codebase

VeckoTheGecko (changes requested):
Good progress! Left some review comments. I think there is some duplication of information between the comments and the code. I recommend in general: [Don't Write Comments - rewrite the code](https://www.youtube.com/watch?v=Bf7vDBBOBUA) (really good YouTube channel in general for code style - highly recommend). The only time I write comments is to explain _why_ something was done. E.g., `# Keeping this legacy format for backward compatibility with older datasets`

There are many ways to make things self documenting. E.g., 
```python
mfp_to_yaml(
            mfp_file, str(path)
        )  # Pass the path to save in the correct directory
```

could become
```python
mfp_to_yaml(
            mfp_file, save_directory = str(path)
        )
```
.

It's possible to have clean, readable code with 0 duplication, but also 0 comments ;)

src/virtualship/expedition/space_time_region.py
```diff
@@ -38,20 +38,19 @@ def _check_lon_lat_domain(self) -> Self:
...
+    end_time: datetime | None = None
 
     @model_validator(mode="after")
     def _check_time_range(self) -> Self:
-        if not self.start_time < self.end_time:
-            raise ValueError("start_time must be before end_time")
+        if self.start_time and self.end_time:
+            if not self.start_time < self.end_time:
+                raise ValueError("start_time must be before end_time")
         return self
```
iuryt, line 53: @VeckoTheGecko @ammedd 
While using pydantic, I had to change this to accept nonetype time, but this means that when we do `schedule.from_yaml()` for `fetch` it will not give any error, right? What should we do?

iuryt:
@ammedd 
I tested the code, and it seems to be working on my end, including the warning and error messages. I also ran the unit tests, which appear to be functioning properly. Let me know if I should add any other functionalities. I hope this is sufficient for you to merge so you don't have a lot of work tomorrow.

tests/test_mfp_to_yaml.py
```diff
@@ -0,0 +1,102 @@
...
+from virtualship.utils import mfp_to_yaml
+
+# Sample correct MFP data
+VALID_MFP_DATA = pd.DataFrame(
+    {
+        "Station Type": ["A", "B", "C"],
+        "Name": ["Station1", "Station2", "Station3"],
+        "Latitude": [30, 31, 32],
+        "Longitude": [-44, -45, -46],
+        "Instrument": ["CTD, DRIFTER", "ARGO_FLOAT", "XBT, CTD, DRIFTER"],
```
ammedd, line 17: Do we need to mention each of the separate instruments here?

VeckoTheGecko: just a test :)

src/virtualship/expedition/space_time_region.py
VeckoTheGecko: I see. I thought that pydantic would have a way of disabling validation during the initialisation of the object, but looking further at the documentation its looking that that isn't possible ... . Longterm it would be good to have `start_time` and `end_time` not none, but perhaps thats something for a future PR

src/virtualship/expedition/waypoint.py
```diff
@@ -1,16 +1,28 @@
...
+    def serialize_instrument(self, instrument):
+        """Ensure InstrumentType is serialized as a string (or list of strings)."""
+        if isinstance(instrument, list):
+            return [inst.value for inst in instrument]
+        return instrument.value if instrument else None
+
+    @field_serializer("time")
+    def serialize_time(self, time):
+        """Ensure datetime is formatted properly in YAML."""
+        return time.strftime("%Y-%m-%d %H:%M:%S") if time else None
```
VeckoTheGecko, line 28: I'm not 100% sure what this code change is. Is there a need to have specific field serializers for these? I imagine that the default Pydantic model serialization would be sufficient?

Also, the changing from `dataclass` from `BaseModel` changes the init signature requiring things to be passed in via kwargs and not args (which isn't a problem, but just wanting to get a better picture of things before shipping).

Reverting for now, but we can put in a furture PR :)

VeckoTheGecko: i see why this was done now - that InstrumentType serialization wasn't happening as expected. Ok, let's keep this (though still not sure about the time serialization as that looks to be ok?)

ammedd (approved):
Tested with a csv on my side. Works!

src/virtualship/utils.py
```diff
@@ -97,11 +97,11 @@ def mfp_to_yaml(excel_file_path: str, yaml_output_path: str):  # noqa: D417
...
-        "XBT": {"maximum_depth": 2000, "buffer": 1},
-        "CTD": {"maximum_depth": 5000, "buffer": 1},
-        "DRIFTER": {"maximum_depth": 1, "buffer": 5},
-        "ARGO_FLOAT": {"maximum_depth": 2000, "buffer": 5},
+    instrument_max_depths = {
+        "XBT": 2000,
+        "CTD": 5000,
+        "DRIFTER": 1,
+        "ARGO_FLOAT": 2000,
     }
```
iuryt: @VeckoTheGecko 
Wasn't better to keep the buffer? The users will have to add this manually?

VeckoTheGecko: Heyo! Sorry if I forgot to explain this in another message (its been a bit of a hackathon here to get things good to go). We were thinking of having the buffering in the actual data download, and not the specification of the spacetime region (so the spacetime region then becomes the region of interest). This would mean that we can reduce the data footprint by looking on a per-dataset level

cc @ammedd

PR #113, opened 2025-01-29
Add assignment for an ECR expedition proposal

VeckoTheGecko:
Continuation of https://github.com/OceanParcels/virtualship/pull/87

> This assignment is designed for graduate students with little former experience in python and should take 10 to 20 hours to complete.


ammedd:
The folder tutorials now occurs twice. Can we remove the instance directly under `docs` as these and other tutorials can now be found under `user-guide/docs`?

ammedd:
@erikvansebille can you review the proposals? The Virtualship one is the one I propose to use for DyOc this year. The RP_only is the one I used in OSL and that you've reviewed in #87

erikvansebille:
> @erikvansebille can you review the proposals? The Virtualship one is the one I propose to use for DyOc this year. The RP_only is the one I used in OSL and that you've reviewed in #87

Yes will do tomorrow

VeckoTheGecko:
> The folder tutorials now occurs twice. Can we remove the instance directly under `docs` as these and other tutorials can now be found under `user-guide/docs`?

This folder should be empty on your machine, and is safe to delete.

Git doesn't have a concept of folders, which means that empty folders can't be tracked. So effectively, once the last file in a folder is gone the folder is effectively "deleted" according to git. More info: https://www.freecodecamp.org/news/what-is-gitkeep/

docs/user-guide/tutorials/Research_Proposal_only.ipynb
erikvansebille, line 1: Some feedback/suggestions on this notebook
1. Use "student grant" instead of "ECR grant" in the title, as we address the readers as students below
2. Also work with subheadings on the paragraphs after the instrumentation details. E.g. a header "EEZs", "Polar work", "Transport and preparation"?
3. Give a bit of information above "Outline" that the goal is that students will out that part of the form? 
4. At Cruise Period, doe we need to limit that to the availability of CMS data already?
5. I like point 12 and 13!
6. Do we also want to have a question about the number of berths/scientists required to do the work? Or is that too difficult?

docs/user-guide/tutorials/Virtualship_research_proposal.ipynb
erikvansebille, line 1: Partly same comments as above. But if the same information is on this tutorial too, would it not make sense to link to this one in the other; so that all information is only in one place (and we avoid mismatches and/or having to update two texts)?

ammedd: Yes, it would be great not to have to update two texts, but I am unsure on how to go about this. Would you suggest making a third notebook with the similar part that both other notebooks then refer to? E.g. 
```
Create two notebooks, A. ipynb and B. ipynb, in the same directory.
In A. ipynb, create a markdown cell with the following contents: [Other notebook](./B. ipynb) .
Render the markdown cell in A. ipynb. Clicking on the "Other notebook" link should open B. ipynb in JupyterLab.
```

docs/user-guide/tutorials/Research_Proposal_only.ipynb
ammedd: 1 - 3: Will do
4: Not for the proposal only (if they are not going to run the code) 
5: :-)
6: With only physical oceanography measurements available I feel it is not needed yet. I was nice to see how creative the students were with it though. Some already made a long list of participants for example.

docs/user-guide/tutorials/Virtualship_research_proposal.ipynb
erikvansebille: Not sure if you can load one notebook into another. But you could otherwise have three notebooks, where you link to each otters at the beginning or the end? So not in one view, but more as chapters? Does that make sense?

ammedd: I'll give it a try

ammedd: Can you check @erikvansebille?

erikvansebille: Yes, this could work, although I don't see the background text anymore? 

And I would move the Introduction to _before_ the Research Proposal for ECR Grants. Perhaps also rename the Introduction on the Assignments Overview list? https://virtualship--113.org.readthedocs.build/en/113/user-guide/tutorials/index.html

And add thumbnails for the two Assignments on that list that don't have one? See https://nbsphinx.readthedocs.io/en/0.7.1/gallery/thumbnail-from-conf-py.html for how to do that

ammedd: Better?

erikvansebille:
Yep, looks great! Only one suggestion: to put `Assignments` as a separate heading in https://virtualship--113.org.readthedocs.build/en/113/user-guide/index.html, because it's now confusing that the Assignments are under Tutorials?

erikvansebille:
Almost done; but the Tutorials folder now appear twice at https://virtualship--113.org.readthedocs.build/en/113/user-guide/index.html and the Assignments don't have thumbnails anymore ;-)

erikvansebille (approved):

ammedd:
Jeej! Shall I next upload the rubrics? For sci com as well?

Issue #116, opened 2025-02-05
Small issues to discuss/fix


ammedd:
1. simulate_schedule.py l.251 time hardcoded
`time_costs.append(timedelta(minutes=20))`
Should come from ship_config.yaml l.16
`  stationkeeping_time_minutes: 20.0`

2. Would it be nice to set time delta and output for drifters and Argos? simulate_measurements l.87 and l.102
`            outputdt=timedelta(hours=5),
            dt=timedelta(minutes=5),`

3. Before there was the option to specify "max" as CTD_depth. Now should I advice students to set a depth more than their expected bathymetry? CTD.py l.27 and l.90
`Variable("max_depth", dtype=np.float32),` 
Enhancement: Would be nice to set this separately for each CTD? And Argo drift_depth and drifter depth_meter as well in schedule.yaml. Then if not specified will be overwritten by ship_config?

4. XBT added to instruments but not to simulate_schedule and simulate_measurements?

5. Change ship speed to kts in config? Recompute to m/s in ship_config l.95

PR #117, opened 2025-02-12
Adds XBTs to schedule and expedition 

ammedd:
- [x] Closes point 4 from #116 
- [ ] Test was created by Iury in #76, need to check if it needs updating and/or an extra test


ammedd:
I don't think we need an extra test, but let me know if you disagree

erikvansebille (approved):

Issue #119, opened 2025-02-13
Fetch only downloads default dataset - add argo and drifter


ammedd:
The fetch command, as defined on `virtualship/src/virtualship/cli/commands.py` line 68 only downloads a default dataset. I'd expect it to check which instruments are used (= stated in `schedule.yaml`) and also download e.g. drifter and argo data if needed.

PR #122, opened 2025-02-14
add drifter + argo data dowload (quick fix, can be made nicer)

ammedd:
Only default data was downloaded, added argo and drifter with buffers around it


ammedd:
Closes #119

PR #127, opened 2025-02-14
speed and other little things

### LINKED ISSUE #116: Small issues to discuss/fix

erikvansebille:
- [x] Closes #116


erikvansebille:
FYI, readthedocs on https://github.com/OceanParcels/virtualship/pull/133 also failed, even though that PR had nothing to do with documentation...

VeckoTheGecko:
Brought this up to date with main (after #137 was merged) and did some cleanup so this avoids doubling up etc.

VeckoTheGecko:
I've separated out the DrifterConfig.period changes into a separate branch (`drifter-period`) to be included in another PR.

- wasn't 100% clear the motivation for it/how it would be used in the codebase/what it's default would be (would there be cases where it would have a default?)
- was causing ci to fail

cc @ammedd

PR #152, opened 2025-02-27
Add support for csv files as input

### LINKED ISSUE #153: Support CSV as an input file format
**Is your feature request related to a problem? Please describe.**
It would be nice to be able to provide the mfp file as a csv file, which is a more general file format than excel.

diogomatoschaves:
- Closes #153 
- [x] Tests added
- [x] Re-ran README.md help commands if the CLI has changed

### DIFF
--- a/src/virtualship/utils.py
+++ b/src/virtualship/utils.py
@@ -45,25 +45,24 @@ def _generic_load_yaml(data: str, model: BaseModel) -> BaseModel:
 
 def load_coordinates(file_path):
     """Loads coordinates from a file based on its extension."""
-
     if not os.path.isfile(file_path):
         raise FileNotFoundError(f"File not found: {file_path}")
 
     ext = os.path.splitext(file_path)[-1].lower()
 
     try:
-        if ext in ['.xls', '.xlsx']:
+        if ext in [".xls", ".xlsx"]:
             return pd.read_excel(file_path)
 
-        if ext == '.csv':
+        if ext == ".csv":
             return pd.read_csv(file_path)
 
         raise ValueError(f"Unsupported file extension {ext}.")
 
     except Exception as e:
         raise RuntimeError(
-            f"Could not read coordinates data from the provided file. "
-            f"Ensure it is either a csv or excel file."
+            "Could not read coordinates data from the provided file. "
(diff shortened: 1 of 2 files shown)


src/virtualship/utils.py
```diff
@@ -42,7 +42,25 @@ def _generic_load_yaml(data: str, model: BaseModel) -> BaseModel:
...
+
+    try:
+        return pd.read_csv(file_path)  # Try reading as CSV with commas
+    except ValueError:
+        pass
+
+    df = pd.read_csv(file_path, delimiter=";")  # Try reading CSV with colons
+    if len(df.columns) == 1:
+        return None
+    return df
```
iuryt, line 60: Since the file_path will contain the extension, isn't is better to just infer the file type by the extension instead of try and error?

iuryt:
Thanks for contributing!

src/virtualship/utils.py
```diff
@@ -42,37 +43,33 @@ def _generic_load_yaml(data: str, model: BaseModel) -> BaseModel:
...
-    from virtualship.expedition.waypoint import Location, Waypoint
+        if ext == ".csv":
+            return pd.read_csv(file_path)
+
+        raise ValueError(f"Unsupported file extension {ext}.")
 
+    except Exception as e:
+        raise RuntimeError(
+            "Could not read coordinates data from the provided file. "
+            "Ensure it is either a csv or excel valid file."
```
erikvansebille, line 65: Not sure what "valid" means in this context...
```suggestion
            "Ensure it is either a csv or excel file."
```

src/virtualship/utils.py
diogomatoschaves: The idea was that sometimes a file can have a missing extension while still being valid, but I see your points 🙌

I have changed the implementation to check based on the file extension, and also improved the error handling.

diogomatoschaves:
Hi @erikvansebille, I made some changes as requested—thanks again for your feedback! Let me know if you have any further comments on this updated version.

VeckoTheGecko (approved):

VeckoTheGecko:
Looks good. Thanks for the contribution @diogomatoschaves !

PR #158, opened 2025-02-28
Fixing bugs where ship went too fast

erikvansebille:
This PR fixes two bugs that make the ship finish too quickly.
1. The conversion from kots/hr to m/s was the wrong way around, leading to ~4x higher ship speeds
2. The ship wouldn't wait at the waypoint until it was planned to be there, if it arrived too early


ammedd (approved):
Thanks!

PR #163, opened 2025-03-06
Add schedule verification step to fetch command

### LINKED ISSUE #162: Run verify_schedule at start of fetch command
> @erikvansebille I would also add that, since the `fetch` step uses the schedule to download the relevant data based on the start and end dates of the trip, the `verify_schedule` function should happen as the first step of `fetch`, and not down the line on the `run` step. 

 _Originally posted by @diogomatoschaves in [#144](https://github.com/OceanParcels/virtualship/issues/144#issuecomment-2694151487)_

diogomatoschaves:
- Closes #162

- Moves `_get_schedule` and `_get_ship_config` to utils file.
- Refactors validation of `schedule` and `ship_config`, which become methods of the `Schedule` and `ShipConfig` classes respectively.
- Makes `get_instruments_in_schedule` a method of the `Schedule` class
- Adds more tests for the `Schedule` and `ShipConfig` classes.
- Moves fetch function to its own module
- __Adds schedule verification step at the beginning of the fetch command__ 

- [x] Tests added


diogomatoschaves:
Hi reviewers, I'm aware that this PR changes more things than the [issue](#162) it is associated with, but I felt some reorganizing of the code was in order as I was making the changes. If it's too much (in particular the moving of the `fetch` command function to another module), let me know and I can revert that back. 

The commits follow a coherent chronological order, and should be viewed individually to make sense of the changes.

tests/cli/test_fetch.py
```diff
@@ -15,6 +16,50 @@
...
+
+    ship_config = ShipConfig.from_yaml(out_path)
+
+    return ship_config
+
+
+@pytest.mark.usefixtures("copernicus_subset_no_download")
+def test_fetch(schedule, ship_config, tmpdir):
+    """Test the fetch command, but mock the download."""
+    _fetch(Path(tmpdir), "test", "test")
```
VeckoTheGecko, line 62: This seems to be a double-up on `test_fetch` in `tests/cli/test_cli.py`? Thoughts? The `runner` fixture I think is also a better way to handle the setting up of these resources.

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+@pytest.fixture
+def ship_config(tmpdir):
+    out_path = tmpdir.join("ship_config.yaml")
+
+    with open(out_path, "w") as file:
+        file.write(get_example_config())
+
+    ship_config = ShipConfig.from_yaml(out_path)
+
+    return ship_config
```
VeckoTheGecko, line 55:
```suggestion
@pytest.fixture
def ship_config(tmp_path):
    tmp_path.write(get_example_config())
    return ShipConfig.from_yaml(out_path)
```

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+    out_path = tmpdir.join("ship_config.yaml")
+
+    with open(out_path, "w") as file:
+        file.write(get_example_config())
+
+    ship_config = ShipConfig.from_yaml(out_path)
+
+    delattr(ship_config, "xbt_config")
+
+    return ship_config
```
VeckoTheGecko, line 69:
```suggestion
@pytest.fixture
def ship_config_no_xbt(ship_config):
    delattr(ship_config, "xbt_config")
    return ship_config
```

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+    out_path = tmpdir.join("ship_config.yaml")
+
+    with open(out_path, "w") as file:
+        file.write(get_example_config())
+
+    ship_config = ShipConfig.from_yaml(out_path)
+
+    delattr(ship_config, "ctd_config")
+
+    return ship_config
```
VeckoTheGecko, line 83:
```suggestion
@pytest.fixture
def ship_config_no_ctd(ship_config):
    delattr(ship_config, "ctd_config")
    return ship_config
```

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+    out_path = tmpdir.join("ship_config.yaml")
+
+    with open(out_path, "w") as file:
+        file.write(get_example_config())
+
+    ship_config = ShipConfig.from_yaml(out_path)
+
+    delattr(ship_config, "argo_float_config")
+
+    return ship_config
```
VeckoTheGecko, line 97:
```suggestion
@pytest.fixture
def ship_config_no_argo_float(ship_config):
    delattr(ship_config, "argo_float_config")
    return ship_config
```

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+    with open(out_path, "w") as file:
+        file.write(get_example_config())
+
+    ship_config = ShipConfig.from_yaml(out_path)
+
+    print(ship_config)
+
+    delattr(ship_config, "drifter_config")
+
+    return ship_config
```
VeckoTheGecko, line 113:
```suggestion
@pytest.fixture
def ship_config_no_drifter(ship_config):
    delattr(ship_config, "drifter_config")
    return ship_config
```

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def test_import_export_ship_config(ship_config, tmpdir) -> None:
+    out_path = tmpdir.join("ship_config.yaml")
+
+    ship_config.to_yaml(out_path)
+
+    shio_config_2 = ShipConfig.from_yaml(out_path)
+    assert ship_config == shio_config_2
```
VeckoTheGecko, line 122:
```suggestion
def test_import_export_ship_config(ship_config, tmp_path) -> None:
    ship_config.to_yaml(tmp_path)

    ship_config_2 = ShipConfig.from_yaml(out_path)
    assert ship_config == ship_config_2
```

tests/expedition/test_ship_config.py
```diff
@@ -0,0 +1,168 @@
...
+@pytest.fixture
+def schedule(tmpdir):
+    out_path = tmpdir.join("schedule.yaml")
+
+    with open(out_path, "w") as file:
+        file.write(get_example_schedule())
+
+    schedule = Schedule.from_yaml(out_path)
+
+    return schedule
```
VeckoTheGecko, line 21: I think we can do the same for schedule as the suggestions below with ship_config.

VeckoTheGecko (changes requested):
Thanks for the contribution. Just have some changes, mainly around the fixtures/tests and function signatures/readability

tests/cli/test_fetch.py
diogomatoschaves: My line of thought is that it makes sense to separate the testing of the fetch function from the testing of the cli command functionality. The latter should focus on the ability of the user to pass some arguments and call the respective function—in this case the `_fetch` function—while the former is testing the actual functionality of what happens inside the `_fetch` function.

tests/expedition/test_ship_config.py
diogomatoschaves: I don't think your suggestion works, as `tmp_path` is not a file name.

src/virtualship/expedition/schedule.py
```diff
@@ -44,3 +52,175 @@ def from_yaml(cls, file_path: str | Path) -> Schedule:
         with open(file_path) as file:
             data = yaml.safe_load(file)
         return Schedule(**data)
+
+    def get_instruments(self, return_objects=False) -> set:
```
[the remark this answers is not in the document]
diogomatoschaves: Yeah agreed, I've changed it accordingly.

diogomatoschaves:
Hey @VeckoTheGecko I was a bit busy the last 2 weeks, so only now making the requested changes. But here they are! Thanks for the review 👌

tests/cli/test_fetch.py
VeckoTheGecko: I see, and would agree if the current implementation of fetch wasn't simply

```py
def fetch(path: str | Path, username: str | None, password: str | None) -> None:
    """
    Download input data for an expedition.

    Entrypoint for the tool to download data based on space-time region provided in the
    schedule file. Data is downloaded from Copernicus Marine, credentials for which can be
    obtained via registration: https://data.marine.copernicus.eu/register . Credentials can
    be provided on prompt, via command line arguments, or via a YAML config file. Run
    `virtualship fetch` on a expedition for more info.
    """
    _fetch(path, username, password)

```

so it feels that they are one and the same. I'll remove this test case for the timebeing, and if the `fetch` command grows (/is refactored) to be more than just `_fetch` then we can have more testing of the individual parts.

VeckoTheGecko (approved):
All good. Thanks for getting back to this @diogomatoschaves ! Looks good to me now :)

erikvansebille (approved):

diogomatoschaves:
There was an error in the tests that had slipped through the cracks. Hopefully now all tests should pass

PR #175, opened 2025-05-01
Prevent fetch from downloading unnecessary data and skipping data download for certain instruments

### LINKED ISSUE #174: Fetch skips downloading data for some instruments
It appears that `virtualship fetch` will only download data from Copernicus Marine Service by virtue of the fact that `“ship_underwater_st_config”` and `“adcp_config”` are specified in `ShipConfig` objects as standard. Meanwhile, the presence of other instruments e.g. `"DRIFTER"` and `"ARGO_FLOAT"` in `schedule.yaml` does not actually prompt a data download. I notice this is related to #122 which has been closed, but I think there could still be problems in the current implementation?

To elaborate, I think there are two problems here:

**1)**  `_fetch()` will *always* download Bathymetry, T, S and U & V by always satisfying the following `if` statement in https://github.com/OceanParcels/virtualship/blob/main/src/virtualship/cli/_fetch.py:

```python
    if (
        (
            {"XBT", "CTD", "SHIP_UNDERWATER_ST"}
            & set(instrument.name for instrument in instruments_in_schedule)
        )
        or hasattr(ship_config, "ship_underwater_st_config")
        or hasattr(ship_config, "adcp_config")
    ):
```

This `if` statement always passes as `True` because `hasattr(ship_config, "ship_underwater_st_config")` and `hasattr(ship_config, "adcp_config")` are always `True`, I think because they are always present as attributes in the `ShipConfig` object (even if set to `None`).

**2)** The`if` statement lines to download data for `"DRIFTER"` and `"ARGO_FLOAT"` added in #122 (e.g.`if "DRIFTER" in instruments_in_schedule:`) appear to always result in `False`. The string doesn't seem to ever be found in `instruments_in_schedule`. It's not immediately clear to me why this is the case because it looks like it should on first glance, but I will look into it. 

**In summary...**

**1)** is currently more of an 'inefficiency' as it's not necessarily stopping anything else from working but rather always downloading data when it's not necessarily always needed. Perhaps `hasattr()` could be replaced by something which means the statement is `False` if `"ship_underwater_st_config"` or `"adcp_config"` are `None` in `ShipConfig` objects?

**2)** is a bug which is preventing the download of extra data such as biogeochemistry so I will look into fixing this before I can carry on with incorporating biogeochemistry workflows as per #171. I suspect this hasn't been an issue up to now because VirtualShip has so far only been used for physical variables which are always downloaded due to what's been outlined in 1) and the data is therefore available for `DRIFTER` and `ARGO_FLOAT` instruments anyway by the time it comes to running `virtualship run`.

j-atkins:
Closes #174

### DIFF
--- a/src/virtualship/cli/_fetch.py
+++ b/src/virtualship/cli/_fetch.py
@@ -88,10 +88,10 @@ def _fetch(path: str | Path, username: str | None, password: str | None) -> None
             {"XBT", "CTD", "SHIP_UNDERWATER_ST"}
             & set(instrument.name for instrument in instruments_in_schedule)
         )
-        or hasattr(ship_config, "ship_underwater_st_config")
-        or hasattr(ship_config, "adcp_config")
+        or ship_config.ship_underwater_st_config is not None
+        or ship_config.adcp_config is not None
     ):
-        print("Ship data will be downloaded")
+        print("Ship data will be downloaded. Please wait...")
 
         # Define all ship datasets to download, including bathymetry
         download_dict = {
@@ -145,8 +145,8 @@ def _fetch(path: str | Path, username: str | None, password: str | None) -> None
         complete_download(download_folder)
         click.echo("Ship data download based on space-time region completed.")
 
-    if "DRIFTER" in instruments_in_schedule:
-        print("Drifter data will be downloaded")
+    if "DRIFTER" in str(instruments_in_schedule):
+        print("Drifter data will be downloaded. Please wait...")
         drifter_download_dict = {
             "UVdata": {
                 "dataset_id": "cmems_mod_glo_phy-cur_anfc_0.083deg_PT6H-i",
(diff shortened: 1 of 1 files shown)


src/virtualship/cli/_fetch.py
```diff
@@ -145,8 +145,8 @@ def _fetch(path: str | Path, username: str | None, password: str | None) -> None
         complete_download(download_folder)
         click.echo("Ship data download based on space-time region completed.")
 
-    if "DRIFTER" in instruments_in_schedule:
-        print("Drifter data will be downloaded")
+    if "DRIFTER" in str(instruments_in_schedule):
```
VeckoTheGecko, line 148: Just wondering what the motivation for this is? Was it not working before?

Since `instruments_in_schedule` is a call to `schedule.get_instruments()` giving a `set[InstrumentType]`, I think the right way to do this would be

```suggestion
    if InstrumentType.DRIFTER in instruments_in_schedule:
```
(also making sure `InstrumentType` is imported at the top)

VeckoTheGecko (changes requested):
Just a small change so that we aren't comparing against a repr :)

src/virtualship/cli/_fetch.py
j-atkins: Yes, much more sensible. I've made these changes. I've added the `InstrumentType` import within the `_fetch` function to avoid circular import errors. Does that seem sensible?

VeckoTheGecko: Yes, that sounds good. I'm not sure why we're getting circular imports, might be worth me looking at the package structure.

But yes, a lot of code sometimes puts the import in the function body rather than the top of the file

Issue #171, opened 2025-04-28
Biogeochemistry


j-atkins:
I will start to add biogeochemistry sampling to VirtualShip. Anticipated first steps will be to:

**1)** add new Field(s) using biogeochem data from Copernicus Marine Service. More specifically, the [Global Ocean Biogeochemistry Analysis and Forecast](https://data.marine.copernicus.eu/product/GLOBAL_ANALYSISFORECAST_BGC_001_028/description) dataset, which is forced by the same Physics Analysis and Forecast already implemented for sampling of physical variables in VirtualShip.

**2)** add oxygen and phytoplankton sampling capabilities onto the existing CTDs.

Extra considerations:
- biogeochem data is distributed on 1/4 degree grid (physics is 1/12 degree and 'coarsened' to force the biogeochem model). Will this require an extra interpolation step in VirtualShip?

PR #183, opened 2025-05-09
New biogeochemistry features

j-atkins:
As part of #171, starting to add new biogeochemistry features. At this point this includes adding bgc data download option (O2 and chlorophyll fields added) when the new 'CTD_BGC' instrument is specified in the scheduling. Includes some commits to get branch up to date ad resolve conflicts.


j-atkins:
Just to add, I have made further changes/additions in #184 which defines and implements the new CTD_BGC instrument, as a branch/new PR to this current PR (#183).

src/virtualship/cli/_fetch.py
```diff
@@ -236,6 +238,50 @@ def _fetch(path: str | Path, username: str | None, password: str | None) -> None
...
+                    overwrite=True,
+                    coordinates_selection_method="outside",
+                )
+        except InvalidUsernameOrPassword as e:
+            shutil.rmtree(download_folder)
+            raise e
+
+        complete_download(download_folder)
+        click.echo("CTD_BGC data download based on space-time region completed.")
+
```
VeckoTheGecko, line 284: Hm, I think that this block and the blocks above will need to be refactored at some point since there is quite some duplication. this can be done in a future PR but just something to be aware of I guess.

src/virtualship/cli/_fetch.py
```diff
@@ -236,6 +238,50 @@ def _fetch(path: str | Path, username: str | None, password: str | None) -> None
...
+                    username=username,
+                    password=password,
+                    overwrite=True,
+                    coordinates_selection_method="outside",
+                )
+        except InvalidUsernameOrPassword as e:
+            shutil.rmtree(download_folder)
+            raise e
+
+        complete_download(download_folder)
```
VeckoTheGecko, line 282: I think `complete_download(...)` should only be done at the complete end of the function. Can you remove all the other calls to it in the download blocks and just put it at the end?

VeckoTheGecko (changes requested):
Just some small changes here re. 'completing the download'

I will need to reconsider how some of the data downloading is done here and also how it is testing - not 100% happy with the current implementation (but that can be left to a future PR I think).

VeckoTheGecko (approved):

Issue #155, opened 2025-02-28
Reduce verbose-level of adcp and underwater


erikvansebille:
Currently, the virtualship prints that it is writing to file for every execution, which for the adcs and underwater instruments can quickly flood the log. Better to print this only once (although might require changing in Parcels?)

PR #185, opened 2025-05-15
Log filter for reducing verbose-level of adcp and underwater

j-atkins:
Relates to bug identified in #155 

I've had a quick go at a bit of a hack to avoid the log being flooded with writing-to-file messages for every execution of adcp and underwater measurements. As I say it's a bit of a hack but it avoids having to change anything in Parcels, rather it simply prevents duplicate messages from being printed when running virtualship (but only for the adcp and underwater instruments and then it's switched off again).

The way I've written it means the log filter is applied by default for adcp and underwater instruments, but there is an option to switch it off in simulate_measurements.py by setting `log_filter` to `False` in `simulate_adcp`/`simulate_ship_underwater_st`. 

Also, not sure if it's a bit messy to have log_filter.py sitting in src/virtualship/, is there somewhere better to house it?

@erikvansebille @VeckoTheGecko @ammedd what do you think?

### DIFF
--- a/src/virtualship/instruments/adcp.py
+++ b/src/virtualship/instruments/adcp.py
@@ -1,10 +1,12 @@
 """ADCP instrument."""
 
+import logging
 from pathlib import Path
 
 import numpy as np
 from parcels import FieldSet, ParticleSet, ScipyParticle, Variable
 
+from ..log_filter import DuplicateFilter
 from ..spacetime import Spacetime
 
 # we specifically use ScipyParticle because we have many small calls to execute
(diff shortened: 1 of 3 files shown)


erikvansebille:
Thanks for implementing this, @j-atkins. I agree that adding a filter to avoid many duplicate messages is good, but I leave it to @VeckoTheGecko to see if this is the most effective/cleanest way to implement it

VeckoTheGecko:
I was talking with Jamie about this yesterday during our 1on1, and settled that it would be nice to suppress the Parcels messages here and to propogate our own. It will be cleaner - and decouples this from Parcels dev

j-atkins:
Thanks both! I will need to make some changes to my original PR to instead suppress all Parcels messages (not just duplicates) and propagate our own. I'll put this on my to do list.

Issue #186, opened 2025-05-15
Features to improve user experience


j-atkins:
After various discussions with @ammedd and @VeckoTheGecko recently, I think there are a few features which could be useful for improving the user experience (in addition to the `animate` idea in progress in #173!). Especially for if and when virtualship is used by people with less scientific/coding experience. 

A few which come to mind:

1) I know there's been previous discussion about the `init`/expedition schedule creation procedure and minimising how much students have to manually input to .xslx files to minimise potential for user error arising from that. However, I've also heard that students don't find manually entering details to the .yaml file intuitive. Is it worth revisiting this? Could we make some kind of simple tool where the .xslx export from MFP is read in, a simple form is launched where per sampling site there are tick boxes for the different instruments, dates and times entered in a little calendar, and then the completed form is automatically written to another .xslx/csv to be used by `virtualship init`? Rather than manually typing these details in the .yaml?

2) Some kind of visual cue that the measurements are running in `virtualship run`. Progress bars are a bit tricky because some measurements are quick (e.g. CTDs) and I imagine a progress bar would just do big % jumps if there were only a few deployments. I understand that's the reason for only having the Parcels `verbose_progress = True` for the adcp and drifter instruments, which take longer (i.e. #133), and not all instruments. I think even just printing a dynamic/rotating loading symbol at the end of the "Simulating {instrument}" log line could be good. Maybe it's just me but I think it's a nice touch to assure that things are underway!

3) A new spin off repo which houses some virtualship post-expedition post-processing tools, e.g. makes a nice dashboard of simple plots for the different instruments and sites. Could be used as a quick visualisation tool for the expedition's data or a starting point for further analysis.

Let me know what you think or if you think there are other user experience features which could be added in the near-medium term future!

erikvansebille:
Thanks for opening this discussion, @j-atkins. These are all three very good and relevant points!  

Some first reflections: 
 
- Point 2 (the verbose output) appears to be the simplest to implement, I guess? So you can go ahead with that if you want
- Point 3 (the post-processing) could also help with integration testing. It would/could provide a way to validate that the output of the virtualship is what we expect? Perhaps, therefore, put it in the same repository as the virtual ship itself? @VeckoTheGecko, what do you think?
[...]

PR #188, opened 2025-05-20
Log filter v2

j-atkins:
This PR is an updated version of #185 (now closed), originally relating to #155. Rather than simply removing duplicate messages in the log, all Parcels messages are now suppressed to decouple from Parcels, as suggested by @VeckoTheGecko  in #185.

I have made further updates to what's printed for the user as they run VirtualShip. This includes adding spinning dials to most instruments to provide a visual cue that the measurement simulation is running (point 2 in #186). For the instruments which take longer to simulate (drifters and argo floats) I have left the call to use Parcels `verbose_progress = True` to get a progress bar. I would have liked to have decoupled from Parcels again here but it proved tricky to propagate our own progress bar given the iteration occurs within the `.execute()` and it is difficult to access this or replicate the iteration outside of the Parcels internals.

Finally, I've reorganised some of the print statements in do_expedition.py to provide a bit more structure to what's printed for the user. It's subjective but hopefully it improves readability and tracking the simulation progress a bit? See an example screenshot below for a completed expedition (the spinning dials are replaced with "[COMPLETED]" when the execution is finished).

Screenshot 2025-05-20 at 16 55 02

The Parcels progress bar doesn't fit in so nicely but maybe something for a future PR...

### DIFF
--- a/src/virtualship/instruments/adcp.py
+++ b/src/virtualship/instruments/adcp.py
@@ -7,6 +7,7 @@ from parcels import FieldSet, ParticleSet, ScipyParticle, Variable
 
 from ..log_filter import Filter, external_logger
 from ..spacetime import Spacetime
+from ..utils import RotatePrint
 
 # we specifically use ScipyParticle because we have many small calls to execute
 # there is some overhead with JITParticle and this ends up being significantly faster
@@ -42,6 +43,8 @@ def simulate_adcp(
     :param num_bins: How many samples to take in the complete range between max_depth and min_depth.
     :param sample_points: The places and times to sample at.
     """
+    rotator = RotatePrint("Simulating onboard ADCP...")
+
     sample_points.sort(key=lambda p: p.time)
 
     bins = np.linspace(max_depth, min_depth, num_bins)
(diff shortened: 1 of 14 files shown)


ammedd:
Nice! I think the water statements are a bit too much too. Perhaps reformulating them to something more general about the route?
`Verifying all waypoints are on water...
... Good, all waypoints are on water.`
E.g. Verifying route... 
... All good to go!

src/virtualship/instruments/adcp.py
```diff
@@ -60,19 +64,33 @@ def simulate_adcp(
...
+                dt=1,
+                runtime=1,
+                verbose_progress=False,
+                output_file=out_file,
+            )
+
+    finally:
+        rotator.stop()
+        for handler in external_logger.handlers:
+            handler.removeFilter(handler.filters[0])
```
VeckoTheGecko, line 96: In Python, this `try` `finally` pattern is normally done via context managers. This means that if something fails in the block of code, the code is torn down after.

This will simplify both the code for the spinner, as well as the code for the filtering, boiling down to something like

```py
with spinner("Deploying drifters..."):
    with discard_parcels_warnings():
        # code to process measurement
```

VeckoTheGecko (changes requested):
Have you considered using https://github.com/pavdmyt/yaspin ? Looks like a good package with options for this. It has minimal dependencies and is quite well maintained.

PR #189, opened 2025-05-22
Move configuration objects to `virtualship.models`

VeckoTheGecko:
This PR introduces a `models` subpackage which is used to house Pydantic models and dataclasses used in the configuration of virtualship. This separation makes it easier to distinguish code for configuration vs running the simulation.

This PR also cleans up the `__init__.py`, cleaning up the mechanism for imports

- xref https://github.com/OceanParcels/virtualship/issues/176


pyproject.toml
```diff
@@ -95,6 +95,7 @@ select = [
     "ICN", # import conventions
     "RUF", # ruff
     "ISC001", # single-line-implicit-string-concatenation
+    "TID", # flake8-tidy-imports
```
VeckoTheGecko, line 98: This promotes more explicit imports (`from virtualship.models` import ...` as opposed to `from ..models import ...`). See [here](https://docs.astral.sh/ruff/rules/#flake8-tidy-imports-tid) for more info

j-atkins (approved):
Looks good to me! I like the restructuring

PR #191, opened 2025-05-27
Updated logging (informative log messages, spinners for progress, suppressing Parcels info messages)

j-atkins:
This PR is an updated and re-imagined version of what I was trying to achieve in #188 (now closed). This version filters out Parcels INFO log messages (to avoid flooding the log) and introduces a yaspin spinner for the output messages. 

Print statements as the expedition runs are also updated to guide the user through what's happening at each step. Tests updated as well.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,6 +13,7 @@ dependencies:
   - pyyaml
   - copernicusmarine >= 2
   - openpyxl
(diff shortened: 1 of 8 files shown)


erikvansebille (approved):
Yep, looks great!

Issue #198, opened 2025-08-22
Copernicusmarine bug


j-atkins:
I've been getting some things ready for an application of VirtualShip in a course at the start of September, and found `virtualship fetch` is consistently failing with:

```
...

  File "/opt/anaconda3/envs/ship/lib/python3.10/site-packages/botocore/client.py", line 1058, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (429) when calling the HeadObject operation: Too Many Requests
```

This same bug has already been [reported](https://github.com/mercator-ocean/copernicus-marine-toolbox/issues/400) in the copernicusmarine issue tracker and appears related to zarr dependencies. There is a [PR](https://github.com/mercator-ocean/copernicus-marine-toolbox/pull/401) with a fix but it's not yet merged and I'm not sure what their timeline will be on that. 

Hopefully the fix should be implemented in time for early September but it leaves me wondering how we should handle updating VirtualShip once it is... I suppose ideally copernicusmarine will do a release and we can update our dependencies. If not, are there options to patch the change into the VirtualShip build ourselves? 

@VeckoTheGecko would you have any ideas here?

cc @ammedd @erikvansebille

VeckoTheGecko:
Good investigation! Indeed looks to be related to Zarr v2. Unfortunately zarr v3 support in Parcels will only arrive in v4 of Parcels.

We can:
- upper-pin copernicus marine so that it gets a lower version (which I think would work better than the latest)
- Wait for the PR to be merged and for a new release to come
- pip install directly from the branch with the fix (i.e., `pip install git+https://github.com/mercator-ocean/copernicus-marine-toolbox@adapting-zarr-v2`)

Patching from the virtualship side will either not be possible or be more effort than its worth

uriii3:
You can try the version 2.2.2, it should avoid the ClientErrors. Let us know if you have any problems related with it!

PR #200, opened 2025-08-27
Update to copernicusmarine v2.2.2

j-atkins:
Updated copernicusmarine dependency to v2.2.2, which fixes the bug raised in #198. 

`virtualship fetch` is now working again!

### DIFF
--- a/tests/make_realistic/test_ctd_make_realistic.py
+++ b/tests/make_realistic/test_ctd_make_realistic.py
@@ -1,12 +1,12 @@
-import seabird
+# import seabird
 
-from virtualship.make_realistic import ctd_make_realistic
+# from virtualship.make_realistic import ctd_make_realistic
 
 
-def test_ctd_make_realistic(tmpdir) -> None:
-    # add noise and convert to cnv
-    files = ctd_make_realistic("ctd.zarr", out_dir=tmpdir, prefix="CTD_")
+# def test_ctd_make_realistic(tmpdir) -> None:
+#     # add noise and convert to cnv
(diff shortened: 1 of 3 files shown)


ammedd:
Yes, please do comment out the seabird package for now. We'll add it later indeed. It requires a bit of thinking/discussing on our side.

tests/make_realistic/test_ctd_make_realistic.py
```diff
@@ -1,12 +1,12 @@
...
-from virtualship.make_realistic import ctd_make_realistic
+# from virtualship.make_realistic import ctd_make_realistic
 
 
-def test_ctd_make_realistic(tmpdir) -> None:
-    # add noise and convert to cnv
-    files = ctd_make_realistic("ctd.zarr", out_dir=tmpdir, prefix="CTD_")
+# def test_ctd_make_realistic(tmpdir) -> None:
+#     # add noise and convert to cnv
+#     files = ctd_make_realistic("ctd.zarr", out_dir=tmpdir, prefix="CTD_")
```
erikvansebille, line 8: I would either add a TODO comment with reference to the GitHub Issue if we expect the package to be back online later; or completely remove (instead of comment out) seabird when we don't expect it to be useable anytime in the future

```suggestion
# TODO readable this test when seabird package works again (see #200)
# def test_ctd_make_realistic(tmpdir) -> None:
#     # add noise and convert to cnv
#     files = ctd_make_realistic("ctd.zarr", out_dir=tmpdir, prefix="CTD_")
```

PR #211, opened 2025-09-29
Adding reanalysis data

### LINKED ISSUE #210: Using Copernicus reanalysis data
After recent discussions, and based on the requirements of different courses at UU, now is a sensible time to move to using Copernicus **Reanalysis** products, rather than purely the **Analysis & Forecast** products which are used at present. This will mean that expeditions can be run back to 1993 and up to present day (defaulting to the analysis product when the scheduled time is within ~2 months of present day or 2 weeks into the future).

The consideration we will have to make here is that the reanalysis products are only at _daily_ resolution, whereas the analysis product (physical variables at least) are available at (6) hourly. 

My concern is just in case users ever want to use VirtualShip for testing intra-daily physical phenomena (note, BGC is daily in all products), diurnal cycles etc. 

With that in mind I suppose we either make clear that we don't support intra-day functionality if we use the reanalysis products throughout (this is probably the simplest option but removes functionality), or indeed we keep the analysis product for as many years as possible and then state somewhere that intra-day physical measurements are only possible in the near-present (2022 onwards).

Keen to hear your thoughts on the issue @ammedd and @erikvansebille!

j-atkins:
This PR expands the range of Copernicus products available to `virtualship fetch`. Reanalysis (and Reanalysis Interim) data are now available for both physical and biogeochemical data, in addition to the existing Analysis & Forecast products.

`fetch` will now automate the dataset selection based on the times in the expedition schedule (now available from 1993 through to two weeks into the future). Following on from #210, all data is now at **daily** resolution*. 

Following on from the [Copernicus Marine documentation](https://help.marine.copernicus.eu/en/articles/4872705-what-are-the-main-differences-between-nearrealtime-and-multiyear-products), I have incorporated the Reanalysis Interim products to maintain consistency between the Reanalysis and Analysis & Forecast products. In summary the three products cover the following periods:

1. **Reanalysis** (or "hindcast" for biogeochemistry):  ~30 years ago to ~5 years ago
2. **Renalysis interim** (or "hindcast interim" for biogeochemistry): ~5 years ago to ~2 months ago
3. **Analysis & Forecast**: ~2 months ago to ~2 weeks in future

*Note also, certain BGC variables (pH, phytoplankton) are only available as monthly products in hindcast and hindcast interim periods.

Given there are now quite a few different combinations of variables, time periods, temporal resolutions etc. I thought it would be helpful to explain this in the documentation, which I have now done ([Additional information.md](https://github.com/OceanParcels/virtualship/blob/0552d3d56362fb3b37aac9a9e12a727ccae35fef/docs/user-guide/additional_information.md)), for transparency on which products are employed when, in case this is ever important to users. There is now a reference to this in the quickstart guide as well. 

To avoid complicating things further, the `CTD_BGC` zooplankton variable is now dropped given its limited temporal range (~2024 onwards). If it's useful to keep this though I can add it back in and add warnings, documentation etc. to be clear that it's limited to this range.

---

Tests for the new functionality also added, plus existing tests for the CLI and `fetch` are updated. 

Closes #210

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -5,7 +5,7 @@
 :hidden:
 
 Home <self>
-User Guide & Documation <user-guide/index>
+User Guide & Documentation <user-guide/index>
 api/index
 contributing/index
 VirtualShip Website <https://virtualship.oceanparcels.org/>


ammedd (approved):
Looks fine to me. Thanks for incorporating the changes.

PR #212, opened 2025-10-03
Apply needs-triage label to incoming issues

VeckoTheGecko:
Same as https://github.com/OceanParcels/Parcels/pull/2296

- Closes None


VeckoTheGecko:
Perhaps we should have a chat sometime about what labels this project can benefit from? (since I think we haven't really editted them from the defaults that github gave)

PR #217, opened 2025-10-17
Unify config files to expedition.yaml

j-atkins:
This PR consolidates the old dual `ship_config.yaml` and `schedule.yaml` config files into one `expedition.yaml` file, as outlined as an objective in the [VirtualShip design doc](https://github.com/Parcels-code/virtualship/blob/main/design-doc.md#configuration-files) for v1-dev.

- [x] Tests added

### DIFF
--- a/src/virtualship/utils.py
+++ b/src/virtualship/utils.py
@@ -114,7 +114,7 @@ def validate_coordinates(coordinates_data):
 
 def mfp_to_yaml(coordinates_file_path: str, yaml_output_path: str):  # noqa: D417
     """
-    Generates an expedition.yaml file with schedule information based on data from MFP excel file. The ship speed and instrument configurations entries in the YAML file are sourced from the static version.
+    Generates an expedition.yaml file with schedule information based on data from MFP excel file. The ship and instrument configurations entries in the YAML file are sourced from the static version.
 
     Parameters
     ----------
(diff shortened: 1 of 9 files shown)


src/virtualship/static/expedition.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - XBT
+      location:
+        latitude: 0.03
+        longitude: 0.03
+      time: 2023-01-01 03:00:00
+    - location:
+        latitude: 0.03
+        longitude: 0.03
+      time: 2023-01-01 03:00:00
+ship_config:
```
erikvansebille: See comment above; could all the items below also be "unnested"?

src/virtualship/utils.py
```diff
@@ -121,7 +114,7 @@ def validate_coordinates(coordinates_data):
 
 def mfp_to_yaml(coordinates_file_path: str, yaml_output_path: str):  # noqa: D417
     """
-    Generates a YAML file with spatial and temporal information based on instrument data from MFP excel file.
+    Generates an expedition.yaml file with spatial and temporal information based on instrument data from MFP excel file. Ship config portion of the YAML file is sourced from static version.
```
erikvansebille:
```suggestion
    Generates an expedition.yaml file with spatial and temporal information based on instrument data from MFP excel file. The instrument and ship config entries in the YAML file are sourced from the static version.
```

docs/user-guide/tutorials/Argo_data_tutorial.ipynb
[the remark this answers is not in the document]
j-atkins: Ah yes, thanks for pointing this out! I must have accidentally re-run the data loading cell when updating the markdown block. Now changed back to as it was before.

erikvansebille:
Very nice improvement! One small suggestion left: you could move `ship_speed` to under the (revived) `ship_config`, which could then (later) also house other ship information like number of berths (if/when relevant for e.g. planning shifts), icestrength class, max days at sea, etc?

j-atkins:
Are you happy for me to merge into v1-dev @erikvansebille?

erikvansebille:
Yep, go for it!

PR #246, opened 2025-11-24
Add Parcels meeting presentation material to developer-content

j-atkins:
Add presentation material (Quarto slides source code and rendered HTML) plus performance testing/plotting scripts to teacher-content in docs.


j-atkins:
I realise it's not ideal uploading all the `Presentation_files` (not least because it is probably the cause of failures on the pre-commit checks)... will need to have a think on how to add presentation material without this and ideally keeping all figures etc. rendering properly!

PR #253, opened 2025-11-28
Random x, y offset for drifter deployments

j-atkins:
This PR adds a small, random offset/noise to the lat, lon release locations for drifters. This means that if users choose to deploy more than one drifter per waypoint they will not all follow the _exact_ same trajectory and there is dispersion. 

- Drifter tests also updated to release a test drifter further within the sample fieldset, to avoid out of bounds errors on release on the edge of the domain.

### DIFF
--- a/tests/instruments/test_drifter.py
+++ b/tests/instruments/test_drifter.py
@@ -4,8 +4,8 @@ import datetime
 
 import numpy as np
 import xarray as xr
-
 from parcels import FieldSet
+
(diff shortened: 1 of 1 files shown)


j-atkins:
> Should there be a lower limit to the noise as well? To make sure drifters are in effect deployed into another model cell? Or is that not needed due to interpolation?

Indeed not required because of the interpolation.

PR #264, opened 2025-12-09
Remove spacetime-region

### LINKED ISSUE #259: `SpaceTimeRegion` problems when changing waypoints after initialisation
In class, there were frequent out-of-bounds errors in `virtualship run` simulations associated with waypoints outside of the `SpaceTimeRegion` prescribed in `expedition.yaml`. These can be easily remedied by manually adjusting the space-time region, but would ideally be avoided in the first place.

The issue most likely arises when there are changes to waypoint lat/lons (or new waypoints) after the point of initialising routes from MFP. The `virtualship plan` tool will automatically update the spatial range component of the `SpaceTimeRegion` if waypoints are changed to outside the current domain. However, there are no checks for this if directly editing the `expedition.yaml`. 

I see some potential solutions, including:

1. Introduce checks as part of `schedule.verify()`
2. Remove `SpaceTimeRegion` entirely. Domain boundaries in the simulation, when streaming data via `copernicusmarine`, could then be determined from the extremes of the waypoints (+ instrument dependent buffers).
    - This is then also likely more aligned with Parcels v4 integration (#232) where there should be no need for prescribing domain limits.

j-atkins:
### Overview

This PR removes `SpaceTimeRegion` from the expedition config. Now, instead, space-time boundaries are defined in reference to the extremes of the waypoints when building fieldsets (i.e. the minimum/maximum latitude/longitude/time across all waypoints). This avoids divergence between user defined waypoints and the prescribed space-time region (i.e. #259).

I also experimented with not defining any spatial limits at all for the fieldset generation. For instruments which require data at depth (Argos, CTDs, XBTs, ADCP) this is still too slow/memory intensive, so I have kept the lat, lon subsetting in `copernicusmarine.open_dataset(..)`. However, for surface-only instruments (Drifters), the lat, lon subsetting can be relaxed. This is a big advantage because it removes out-of-bounds issues for Drifters. The bathymetry fields also require no lat, lon bounds.

Otherwise of interest - and happily - I think there could a speed boost as result of the changes in this PR (with e.g. Argo floats and the underway instruments), but I haven't confirmed this nor diagnosed exactly why this might be happening.

#### Drifters
- The logic for determining the temporal extent for drifter fieldsets has changed. Before there was a generic 41+21 day time buffer added. 
- Now, the drifter lifetime parameter (in `InstrumentConfig`) also controls the endtime of the fieldset (so it will be the last waypoint time + the drifter lifetime). 
- This, in combination the relaxed lat, lon bounds, means that users can set whatever lifetime they wish without hitting out-of-bounds errors. 

#### Argo floats
- Previously Argo float lifetimes were controlled by the temporal extent of the fieldset. 
- There is a now a lifetime parameter added to the Argo configuration (in `InstrumentConfig`), giving more control to the user. This defaults to 3 months.
- Now, similarly to Drifters, the fieldset endtime is controlled by the lifetime parameter, rather than the other way round.
- Because we cannot yet relax the lat, lon bounds for the Argos, it is still possible to hit out-of-bounds errors with Argos. Though, this is not very common and I think less likely than was previously the case with Drifters at the surface.
- Parcels v4 could allow us to relax the lat, lon bounds on **all** instruments, so this may not even be problem for Argos in the future (?)
    - This would allow VirtualShip Argos to match [real-life Argo lifetimes](https://argo.ucsd.edu/how-do-floats-work/technological-innovations/) in theory...

---- 

- Closes #259 
- [x] Tests updated

### DIFF
--- a/src/virtualship/utils.py
+++ b/src/virtualship/utils.py
@@ -13,8 +13,8 @@ from typing import TYPE_CHECKING, Literal, TextIO
 import copernicusmarine
 import numpy as np
 import xarray as xr
-
 from parcels import FieldSet
+
 from virtualship.errors import CopernicusCatalogueError
 
 if TYPE_CHECKING:


src/virtualship/models/expedition.py
```diff
@@ -214,13 +213,26 @@ class ArgoFloatConfig(pydantic.BaseModel):
...
 
     stationkeeping_time: timedelta = pydantic.Field(
         serialization_alias="stationkeeping_time_minutes",
         validation_alias="stationkeeping_time_minutes",
         gt=timedelta(),
     )
 
+    @pydantic.field_serializer("lifetime")
+    def _serialize_lifetime(self, value: timedelta, _info):
+        return value.total_seconds() / 60.0
```
erikvansebille: Again, would this not make more sense in days?

erikvansebille (commented):
Looks good! Great improvement which will reduce annoying outofbounds errors and thereby increase authenticity!

A few comments below

PR #265, opened 2025-12-10
JOSS manuscript

j-atkins:
This PR is for visibility of a draft of the VirtualShip JOSS [manuscript](https://github.com/Parcels-code/virtualship/blob/joss-paper/docs/paper/paper.md), and for the review as co-authors to be organised here.

Feel free to add comments and commits to this PR, or branch off as new PRs for suggested changes.

----

@VeckoTheGecko @iuryt @surgura you have all made substantial contributions to the code. I have listed you as co-authors, alongside @ammedd and @erikvansebille. In the first place, are you happy to be involved with the manuscript? If so, please also let me know what kind of timeline I can expect to get your feedback!

PR #270, opened 2026-01-16
adapt manuscript to new JOSS scope requirements

j-atkins:
Following on from my [comment](https://github.com/Parcels-code/virtualship/pull/265#issuecomment-3759072520) in the original JOSS manuscript PR, here is the adapted manuscript, adapted to the new JOSS requirements.

In practice, it required minimal changes because the content largely addressed the new requirements anyway but some heading titles needed to be changed and an AI disclosure statement added.

cc @erikvansebille @ammedd @VeckoTheGecko @iuryt @surgura


docs/paper/paper.md
erikvansebille: Same comments as on paper.jats (why is it duplicate?)

erikvansebille (commented):
Looks good; just three small comments below

docs/paper/paper.md
j-atkins: Ah this is a result of rendering the PDF. I will remove the PDF and associated build files though given it's not required for submission (only the `paper.md` and `paper.bib` are needed) and was just helpful for drafting.

Issue #258, opened 2025-12-02
More info on editing expedition.yaml directly


ammedd:
As the bespoke planning tool slows down too much when an expedition has >30 or so waypoints, some more information on directly editing ```expedition.yaml``` in e.g. the quick start guide or the yaml itself would be good.

PR #277, opened 2026-02-03
Plan UI usability (adding `expedition.yaml` guidance)

### LINKED ISSUE #267: Investigate plan UI lag
Users in class recently experienced lag using the `virtualship plan` UI with many waypoints. This needs investigation to diagnose the problem and develop a fix. 

This should be conducted both locally and in remote (SURF) JupyterLab settings.

j-atkins:
Previous courses have identified that the `virtualship plan` UI can become cumbersome to use when an expedition has many waypoints (#267). 

I have now looked into this and see that when running locally, having many (e.g. > 30) waypoints does introduce a small lag to the application but it is still usable. However, doing the same in a remote SURF environment results in substantial lag, confirming what was observed in-class. I can do more investigation on this but in the near-term I don't think there's a quick fix, unfortunately.

Therefore, as suggested in #258, this PR adds an additional doc to the tutorials providing guidance on the structure of the `expedition.yaml` file and how to edit it. There are now also references to this tutorial in the quickstart guide, the sail_the_ship assignment and the YAML itself, encouraging users to interact directly with the YAML file if their expeditions are large.

---

Also includes some small tweaks to the tutorials section layout (and thumbnails) in the docs.

---

Closes #267

### DIFF
--- a/docs/user-guide/tutorials/working_with_expedition_yaml.md
+++ b/docs/user-guide/tutorials/working_with_expedition_yaml.md
@@ -94,12 +94,12 @@ instruments_config:
   ship_underwater_st_config: null
 ```
 
-For **all other instruments**, e.g. CTD, ARGO_FLOAT etc., the paramters can often be left as the default values unless advanced customisations are required.
+For **all other instruments**, e.g. CTD, ARGO_FLOAT etc., the parameters can often be left as the default values unless advanced customisations are required.
 
 #### 3. `ship_config`
 
 This section contains setting related to the ship itself, specifically:
 
-- **Ship speed (`ship_speed_knots`)**: The speed of the ship in knots. Note in most cases this should be left as the default value unless there is a specific reason to change it.
+- **Ship speed (`ship_speed_knots`)**: The speed of the ship in knots (nautical miles per hour; where 1 knot = 1.852 km/h). Note in most cases this should be left as the default value unless there is a specific reason to change it.
 
 <!-- TODO: more details on ship configuration if added in the future -->


docs/user-guide/assignments/Sail_the_ship.ipynb
```diff
@@ -156,6 +156,11 @@
     "\n",
     "The next step is to finalise the expedition schedule plan, including setting times and instrument selection choices for each waypoint, as well as configuring the ship (including any underway measurement instruments). \n",
     "\n",
+    "<div class=\"alert alert-block alert-info\"> \n",
+    "**NOTE**: This section describes the process of finalising the expedition schedule and instrument selection using the `virtualship plan` application. This is the recommended way for most users but when expeditions become larger with many waypoints, it can become cumbersome to use the planning tool (note, using VirtualShip in a remote terminal / cloud-based environment can also introduce lag in the user-interface). **In this case, you may prefer to edit the** `expedition.yaml` **file directly (see [here](../tutorials/working_with_expedition_yaml.md) for more details on how to do so)**.\n",
```
ammedd: Change: `This is the recommended way for most users but when expeditions become larger`
to `For expeditions`

ammedd (commented):
Some small suggestions

j-atkins:
Thanks @ammedd, I've incorporated your suggested edits 👍

docs/user-guide/tutorials/working_with_expedition_yaml.md
```diff
@@ -0,0 +1,105 @@
...
+  ship_underwater_st_config: null
+```
+
+For **all other instruments**, e.g. CTD, ARGO_FLOAT etc., the paramters can often be left as the default values unless advanced customisations are required.
+
+#### 3. `ship_config`
+
+This section contains setting related to the ship itself, specifically:
+
+- **Ship speed (`ship_speed_knots`)**: The speed of the ship in knots. Note in most cases this should be left as the default value unless there is a specific reason to change it.
```
erikvansebille:
```suggestion
- **Ship speed (`ship_speed_knots`)**: The speed of the ship in knots (nautical miles per hour; where 1 knot = 1.852 km/h). Note in most cases this should be left as the default value unless there is a specific reason to change it.
```

PR #279, opened 2026-02-05
Generate post-expedition problems report

j-atkins:
This PR adds a step to generate a human-readable record of all problems which occurred across the expedition (including their full message), which could be useful to refer back to when students are making their presentations.

Also tidies up some of the `problems_encountered` directory strucutre.


erikvansebille (changes requested):
Code changes look good to me; but could you also add some unit-tests so that we don’t unintentionally break this functionality later?

PR #294, opened 2026-02-19
Warning on overwriting `results` dir

### LINKED ISSUE #261: Handle overwriting results directory
Sometimes users will encounter an error (e.g. #259 ) in their `virtualship run` simulations. They can then make their fixes and re-run the simulation. However, this will overwrite the `results` directory which already exists with the results up until the point of error, which can be inconvenient when the error occurs late in the expedition.

It would be good to provide an option not to overwrite the existing results, or to build on what already exists.

j-atkins:
This PR adds a warning stage if the `results` dir already exists in the expedition dir. If it does it will give the option to continue and overwrite the previous results, and also provides messaging that `results` should be moved/renamed if the user doesn't want to overwrite it.

Closes #261

### DIFF
--- a/src/virtualship/cli/_run.py
+++ b/src/virtualship/cli/_run.py
@@ -133,7 +133,7 @@ def _run(
 
     # delete and create results directory
     results_dir = expedition_dir.joinpath(RESULTS)
-    _warn_overwrite(results_dir)
+    _warn_overwrite_results_dir(results_dir)
 
     if os.path.exists(results_dir):
         shutil.rmtree(results_dir)


src/virtualship/cli/_run.py
```diff
@@ -284,6 +288,21 @@ def _unique_id(expedition: Expedition, cache_dir: Path) -> str:
     return new_id
 
 
+def _warn_overwrite(results_dir: Path) -> None:
```
erikvansebille: The function name suggests it's generalisable for all warnings on overwrite, but really it's for the results directly only. So I would, for clarity, also name the function that way


```suggestion
def _warn_overwrite_results_dir(results_dir: Path) -> None:
```

src/virtualship/cli/_run.py
```diff
@@ -284,6 +288,21 @@ def _unique_id(expedition: Expedition, cache_dir: Path) -> str:
...
+    if os.path.exists(results_dir):
+        print(
+            f"\nWARNING: The '{results_dir}' directory already exists and will be overwritten. If you want to keep the previous results, please move or rename the '{results_dir}' directory before re-running the expedition.\n"
+        )
+        decision = input(
+            "Do you want to continue and overwrite the existing results? (y/n): "
+        )
+        if decision.lower() != "y":
```
ammedd: Should this be "n"?

PR #299, opened 2026-03-03
Fix ship/underway instrument sampling trajectories

### LINKED ISSUE #298: ADCP out of bounds errors
cc @erikvansebille @ammedd 

As noted in class, the `ADCP` instrument sometimes gets unexpected out-of-bounds errors. I've diagnosed that this is due to a mismatch in the ship route vs. the waypoint locations (likely stemming from the methods in `simulate_schedule`). 

This can be seen in the example expedition below (the same expedition shared by the student group). The underway instruments want to make measurements along the blue track but the fieldset boundaries (black box) are determined by the waypoint locations (red markers), hence there is an out-of-bounds error in this case (at the eastern edge).

Note, this is therefore also likely to be a bug for the underway `UnderwaterST` instrument.

Image

I'll look deeper into the ship routing methods to see why this is happening.

j-atkins:
Following #298 and the issues identified in class, this PR fixes the the path that the ship takes (and therefore the measurement locations for `ADCP` and `Underwater_ST` instruments) to be the actual shortest path between waypoints on the spherical globe.

As a reminder (and summarised in the plot below), the ship was 'missing' waypoints previously by going straight along a bearing, without curving, and this effect was enhanced when there are larger gaps between waypoints. Hence, the ship could overshoot a waypoint and leave the fieldset domain, causing out of bounds errors. Now, the logic is changed so that first an array of all locations on the forward propagation on the globe are accessed by `geod.fwd_intermediate()` and the number of spacings between each waypoint is determined by the distance and time between waypoints and the period associated with `ADCP`/`Underwater_ST` sampling. These are then added to the list storing where to take measurements.

**Summary**:

image

Closes #298

### DIFF
--- a/tests/expedition/test_simulate_schedule.py
+++ b/tests/expedition/test_simulate_schedule.py
@@ -69,7 +69,7 @@ def test_time_in_minutes_in_ship_schedule() -> None:
 
 
 def test_ship_path_inside_domain() -> None:
-    """Test that the ship path (and therefore where underway measurements will be taken) is inside the domain defined by the waypoints (which determines the fieldset bounds)."""
+    """Test that the ship path (here represented by underway ADCP measurement sites) is inside the domain defined by the waypoints (which determines the fieldset bounds)."""
     base_time = datetime.strptime("2022-01-01T00:00:00", "%Y-%m-%dT%H:%M:%S")
 
     projection = pyproj.Geod(ellps="WGS84")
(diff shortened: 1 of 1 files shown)


erikvansebille:
Thanks for adding the tests; these look good! I assume you tested that the old code fails on them?

j-atkins:
> I assume you tested that the old code fails on them?

Oops yes I should have indeed mentioned. The old code fails and the new passes 👍

PR #300, opened 2026-03-03
Handle `instruments: null` at waypoints

j-atkins:
This PR fixes the waypoint instruments handling in `_calc_wp_stationkeeping_time()` to be able to properly handle `instrument: null` at a waypoint in the expedition.yaml.

Convert `expedition.schedule.instrument` to an empty list when it is `null` in YAML. `null` is takenas `NoneType` in Python and cannot be used for the various checks later in the function.

- [x] Tests

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -4,9 +4,9 @@ from pathlib import Path
 import numpy as np
 import pytest
 import xarray as xr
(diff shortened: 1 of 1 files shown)


j-atkins:
> Again, I assume that the test fails on the old code?

Yes, indeed, the previous version fails

PR #324, opened 2026-04-08
New `analysis` pixi environment

j-atkins:
It's useful to have a few extra dependencies when experimenting/testing code changes with VirtualShip expeditions, plus for post-processing results. Therefore, this PR adds a new pixi environment (called `analysis`) to the project with additional dependencies including `plotly`, `jupyterlab`, `ipykernel` etc. 

@VeckoTheGecko is this the best way to go about this kind of thing with pixi?

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -58,6 +58,16 @@ ipykernel = "*"
 [feature.notebooks.tasks]
 tests-notebooks = "pytest --nbval-lax docs/"
 
+[feature.analysis.dependencies]
+jupyterlab = ">=4,<5"
+ipykernel = ">=7,<8"
+plotly = ">=6,<7"
+nbformat = "*"
+ipywidgets = "*"
+
+[feature.analysis.tasks]
+lab = "jupyter lab"
+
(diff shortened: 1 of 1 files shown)


pixi.toml
VeckoTheGecko: you can instead just add the analysis feature here, so that it's available by default. I think that would be more useful (during testing the `test-*` envs will still be there, so it wont affect that)

```suggestion
default = { features = ["test", "notebooks", "typing", "pre-commit", "analysis"] }
```

pixi.toml
```diff
@@ -92,6 +102,7 @@ test-py310 = { features = ["test", "py310"] }
 test-py311 = { features = ["test", "py311"] }
 test-py312 = { features = ["test", "py312"] }
 test-notebooks = { features = ["test", "notebooks"], solve-group = "test" }
+analysis = { features = ["analysis"], solve-group = "analysis" }
```
VeckoTheGecko:
```suggestion
```
as above

j-atkins:
Great, thanks @VeckoTheGecko!

Issue #313, opened 2026-03-25
Add W to ADCP measurements


ammedd:
W (Upward sea water velocity) is only available in the Global Ocean Physics Analysis and Forecast. Nevertheless, would it be nice to add this variable to the ADCP when possible (2020 onwards)?

Issue #312, opened 2026-03-25
Add Salinity to drifter measurements


ammedd:
Currently our drifters only measure T. Add "sensors" for S as well?
Perhaps in a next version when the CTD is also combined and separate sensors will become available?

Issue #260, opened 2025-12-02
Unify `CTD` and `CTD_BGC` to one instrument


j-atkins:
`CTD` and `CTD_BGC` are probably best unified - in reality they are the same instrument.

Open question about whether there should then be more configuration options where different variables can be switched on/off.

Issue #237, opened 2025-11-19
Instrument recipe book


j-atkins:
With instrument logic being refactored and centralised to the `instruments` as of #226, it would really enhance the documentation to include a 'recipe book' to help users add their own instruments to VirtualShip.

j-atkins:
To add to @ammedd's comment, the refactoring of instruments to base and child classes in #226 should streamline the steps required to add a new instrument. At a high level it means adding a new instrument file to the `instruments/` dir and then each gets a Dataclass, a Particle Class, kernel(s) and an Instrument class which has a `.simulate()` method. Its mapping would then need to be added to [types.py](https://github.com/Parcels-code/virtualship/blob/main/src/virtualship/instruments/types.py) and an instrument-specific config added to the `InstrumentsConfig` model in [expedition.py](https://github.com/Parcels-code/virtualship/blob/main/src/virtualship/models/expedition.py).

We're aiming for a set up where minimal changes additional changes are needed in other areas of the codebase (indeed some issues are currently open e.g. #236) but for now there are still some extra steps when adding a new instrument. For example, any new instrument would have to be explicitly added to the workflow in [simulate_schedule.py](https://github.com/Parcels-code/virtualship/blob/main/src/virtualship/expedition/simulate_schedule.py) and [_plan.py](https://github.com/Parcels-code/virtualship/blob/main/src/virtualship/cli/_plan.py).

j-atkins:
I also like @ammedd's mooring suggestion, it sounds like a nice 'gateway' instrument which can easily be adjusted for sensors at different depths etc, and a good simple model to highlight all instrument code components.

@iuryt is your suggestion though that it's called something else in the tutorial to distinguish from what is/will be available already in VirtualShip? I think worst case it would also be fine just to frame it as 'these are the steps that we would take if we pretend that the instrument doesn't already exist in VirtualShip'

iuryt:
I think that pausing until these and the pixi PR are implemented is better. I will also think about ways we could make this work without user having to change the source code.

iuryt:
[...]

PR #328, opened 2026-04-10
Refactor: instruments have sensors

j-atkins:
## Overview

This PR centralises/abstracts instrument sensor/variable sampling logic so that each instrument declares which sensors it carries (e.g. `TEMPERATURE`, `SALINITY`, `VELOCITY`). Users can configure which sensors are active for each instrument in the expedition YAML.

This helps pave the way for easy addition of BGC sensors to the Argo float in a future PR (#234), and consolidation of CTD + CTD_BGC into a single instrument (#260) with a combined sensor allowlist. Also makes it straightforward to add new sensors to any instrument in the future (e.g., #312, #313), and streamlines them process for developers to add new instruments (i.e. #237)

## Major changes

- `sensors.py` in `instruments/` defines the `SensorType` class and per-instrument allowlists (so that there is control over which sensors each instrument supports and users cannot configure unsupported sensors).
- New `SensorConfig` pydantic model and `sensors` field in every instrument config in `expedition.py`.
- New `SensorRegistry` in `utils.py` that maps each `SensorType` to its FieldSet key, Copernicus variable name, category (`phys`/`bgc`), and Parcels particle variable name(s).
    - Per-instrument parcticle classes are now built dynamically at runtime based on which sensors are active, but the fixed/mechanical variables are still hard-coded in the instrument files, e.g. `cycle_phase` for Argo Floats.

## API change

As mentioned above, the instruments config section of the expedition YAML now has a `sensors` list field, where users specify which sensors are active. For example, below the CTD is configured to sample `TEMPERATURE` and `SALINITY`:

```yaml
ctd_config:
  stationkeeping_time_minutes: 50
  min_depth_meter: -11
  max_depth_meter: -2000
  sensors:
    - TEMPERATURE
    - SALINITY
```

- If the `sensors` list is omitted, it default to all valid sensors for that instrument.
- By using allowlists for each instrument, it will not allow non-sensical sensor combinations (e.g. BGC sensors on an ADCP).
- Will also not allow an empty sensor list, at least one sensor must be active. 

## Additional change

- Argo Float sampling kernels have been separated from the vertical-movement kernel, making it easier to add BGC sensors in a future PR.

## Follow-up PRs

- The `plan` CLI tool will need updating to account for sensor configuration options. Currently not broken but doesn't give option to configure sensors. (New issue to be opened)
- Docs update to clearly communicate which sensors each instrument accepts, and how to configure them in either the plan tool or the expedition YAML. (New issue to be opened)
- Merge CTD + CTD_BGC into a single instrument. (#260)
- Add BGC sampling to Argo Floats (#234)

## Tests

- [x] Update existing tests and add new tests to cover new sensor logic


src/virtualship/instruments/ctd_bgc.py
erikvansebille, line 1: There seems to still be a lot of duplication between CTD and CTD_BGC. Was the intention for this PR not to remove that duplication/ Or will that come in a next PR?

src/virtualship/instruments/drifter.py
```diff
@@ -26,17 +31,14 @@ class Drifter:
...
 
-_DrifterParticle = JITParticle.add_variables(
-    [
-        Variable("temperature", dtype=np.float32, initial=np.nan),
-        Variable("has_lifetime", dtype=np.int8),  # bool
-        Variable("age", dtype=np.float32, initial=0.0),
-        Variable("lifetime", dtype=np.float32),
-    ]
-)
+_DRIFTER_FIXED_VARIABLES = [
```
erikvansebille: Not sure whether FIXED is the right term here (and in other instruments). Suggests they are fixed to the instrument. Would NONSENSOR be a better term? Or do you have another term?

src/virtualship/instruments/sensors.py
```diff
@@ -0,0 +1,52 @@
...
+    TEMPERATURE = "TEMPERATURE"
+    SALINITY = "SALINITY"
+    VELOCITY = "VELOCITY"
+    OXYGEN = "OXYGEN"
+    CHLOROPHYLL = "CHLOROPHYLL"
+    NITRATE = "NITRATE"
+    PHOSPHATE = "PHOSPHATE"
+    PH = "PH"
+    PHYTOPLANKTON = "PHYTOPLANKTON"
+    PRIMARY_PRODUCTION = "PRIMARY_PRODUCTION"
```
erikvansebille, line 24: Why is this list needed? Can we do without? Seems like an extra place to change when adding a sensor..

src/virtualship/instruments/ship_underwater_st.py
```diff
@@ -38,12 +40,18 @@ class Underwater_ST:
...
 # define function sampling Temperature
 def _sample_temperature(particle, fieldset, time):
-    particle.T = fieldset.T[time, particle.depth, particle.lat, particle.lon]
+    particle.temperature = fieldset.T[time, particle.depth, particle.lat, particle.lon]
+
+
+_ST_SENSOR_KERNELS: dict[SensorType, callable] = {
+    SensorType.TEMPERATURE: _sample_temperature,
+    SensorType.SALINITY: _sample_salinity,
+}
```
erikvansebille: Why explicitly mention them here, if they are also part of sensors.py? Seems a duplication?

src/virtualship/models/expedition.py
```diff
@@ -10,12 +10,24 @@
 import yaml
 
 from virtualship.errors import InstrumentsConfigError, ScheduleError
+from virtualship.instruments.sensors import (
```
erikvansebille: Why not import all? So that if a new sensor is added, it can automatically also be imported here?

src/virtualship/models/expedition.py
```diff
@@ -271,6 +351,28 @@ def _serialize_period(self, value: timedelta, _info):
...
+
+    @pydantic.field_serializer("sensors")
+    def _serialize_sensors(self, value: list[SensorConfig], _info):
+        return _serialize_sensor_list(value)
+
+    def active_variables(self) -> dict[str, str]:
+        """
+        FieldSet-key → Copernicus-variable mapping for enabled sensors.
+
+        VELOCITY is a special case: one sensor provides two FieldSet variables (U and V).
```
erikvansebille, line 326: But in Parcels, it should sample `fieldset.UV`, which is one Field?

src/virtualship/models/expedition.py
```diff
@@ -307,6 +430,22 @@ class CTD_BGCConfig(pydantic.BaseModel):
...
+            SensorConfig(sensor_type=SensorType.CHLOROPHYLL),
+            SensorConfig(sensor_type=SensorType.NITRATE),
+            SensorConfig(sensor_type=SensorType.PHOSPHATE),
+            SensorConfig(sensor_type=SensorType.PH),
+            SensorConfig(sensor_type=SensorType.PHYTOPLANKTON),
+            SensorConfig(sensor_type=SensorType.PRIMARY_PRODUCTION),
+        ],
+        description=(
+            "Sensors fitted to the BGC CTD. "
+            "Supported: CHLOROPHYLL, NITRATE, OXYGEN, PH, PHOSPHATE, PHYTOPLANKTON, PRIMARY_PRODUCTION. "
```
erikvansebille: Make this list the same order as the list above? Otherwise, more difficult to see whether the list is complete

src/virtualship/models/expedition.py
```diff
@@ -383,6 +578,24 @@ class XBTConfig(pydantic.BaseModel):
...
+        return _check_sensor_compatibility(value, XBT_SUPPORTED_SENSORS, "XBT")
+
+    @pydantic.field_serializer("sensors")
+    def _serialize_sensors(self, value: list[SensorConfig], _info):
+        return _serialize_sensor_list(value)
+
+    def active_variables(self) -> dict[str, str]:
+        """FieldSet-key → Copernicus-variable mapping for enabled sensors."""
+        return build_variables_from_sensors(self.sensors)
+
```
erikvansebille: Why are these functions required for every instrument? Can't they be reused?

src/virtualship/utils.py
```diff
@@ -52,6 +55,98 @@
...
+            particle_vars=["salinity"],
+        ),
+        SensorType.VELOCITY: _SensorMeta(
+            fs_key="UV",
+            copernicus_var="uo",  # uo is primary var here... active_variables() in ADCPConfig expands to both uo and vo
+            category="phys",
+            particle_vars=[
+                "U",
+                "V",
+            ],  # two particle variables associated with one sensor
```
erikvansebille: Why not `UV`?

tests/instruments/test_argo_float.py
erikvansebille, line 1: Great to add so much more unit testing!

erikvansebille (commented):
Looks good! A few comments below

src/virtualship/instruments/ctd_bgc.py
j-atkins: Yes, indeed it will come in the next PR... the CTD_BGC will soon be removed but I decided not to do it in this PR so as to keep it focused

src/virtualship/models/expedition.py
j-atkins: Oops! Overcomplicated things here unnecessarily

src/virtualship/instruments/sensors.py
j-atkins: I prefer to keep it just because it's a lightweight way of adding a bit of consistency/convention across the codebase and helps with some things like quickly catching typos in the YAML config

src/virtualship/utils.py
```diff
@@ -52,6 +55,98 @@
...
 
+# =====================================================
+# SECTION: sensor and variable metadata and registries
+# =====================================================
+
+
+@dataclass(frozen=True)
+class _SensorMeta:
+    fs_key: str  # map to Parcels fieldset variables
+    copernicus_var: str  # map to Copernicus Marine Service variable names
```
VeckoTheGecko: I think it would be good to rename this class to `_Sensor`, and to include the type in the class itself


```suggestion
class _Sensor:
    type_: SensorType
    fs_key: str  # map to Parcels fieldset variables
    copernicus_var: str  # map to Copernicus Marine Service variable names
```

Then the `SENSOR_REGISTRY` can become:

```python
SENSOR_REGISTRY: dict[SensorType, _Sensor] = {s.type_: s for s in [
    ... # list of sensors
]} 
```
making it easy to look up sensors, but also making it so that if a sensor its chosen it's easy to see what type it is again.

src/virtualship/instruments/sensors.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+# per-instrument allowlists of supported sensors (source truth for validation for which sensors each instrument supports)
```
VeckoTheGecko: I find it a bit confusing the structure of the classes here - perhaps this is something worth drawing up so that we have some clarity?

We have an `Instrument` class

```python
@register_instrument(InstrumentType.ADCP)
class ADCPInstrument(Instrument):
    """ADCP instrument class."""

    def __init__(self, expedition, from_data):
        """Initialize ADCPInstrument."""
        variables = expedition.instruments_config.adcp_config.active_variables()
        limit_spec = {
            "spatial": True
        }  # spatial limits; lat/lon constrained to waypoint locations + buffer

        super().__init__(
            expedition,
            variables,
            add_bathymetry=False,
            allow_time_extrapolation=True,
            verbose_progress=False,
            spacetime_buffer_size=None,
            limit_spec=limit_spec,
            from_data=from_data,
        )
```
But I feel that there's a lot of details in this class that aren't really related to what an instrument _is_, but is more glue code to get it working in VirtualShip

From my POV, its best to design things close to the physical domain. An Instrument _instance_ has a set of sensors installed on the machine. These sensors need to match up with a list of allowed sensors (which are defined on the class, e.g., via a method `.get_allowed_sensors()` and a check in the `__init__` that it adheres - this would remove all the `_check_sensor_compatibility` methods that are currently here). Then each sensor has its own details potentially related to that sensor.

Maybe it would be good for us to sit down and diagram out the abstractions here. Part of that might be brainstorming how the interface between the configs and model code looks like.

src/virtualship/utils.py
```diff
@@ -52,6 +55,98 @@
 EXPEDITION_ORIGINAL = "expedition_original.yaml"
 EXPEDITION_LATEST = "expedition_latest.yaml"
 
+# =====================================================
+# SECTION: sensor and variable metadata and registries
+# =====================================================
+
+
+@dataclass(frozen=True)
+class _SensorMeta:
```
VeckoTheGecko: Is it possible for us to also add the kernels to this Sensor class? I see a bunch of mappings across the codebase relating the sensors to the kernels (where a lot of them are the same, as Erik mentioned in https://github.com/Parcels-code/virtualship/pull/328#discussion_r3071188491 ).

I think it would be a big win if we can put the kernels in the sensor class! I don't know how easy it would be since the right kernel might depend on more than just the choice of sensor. If that's the case - what does choosing the right kernel depend on?

src/virtualship/utils.py
```diff
@@ -52,6 +55,98 @@
 EXPEDITION_ORIGINAL = "expedition_original.yaml"
 EXPEDITION_LATEST = "expedition_latest.yaml"
 
+# =====================================================
+# SECTION: sensor and variable metadata and registries
+# =====================================================
```
VeckoTheGecko: Should we move this to `sensors.py`? The class and the `SENSOR_REGISTRY`?

src/virtualship/utils.py
```diff
@@ -52,6 +55,98 @@
...
+
+
+@dataclass(frozen=True)
+class _SensorMeta:
+    fs_key: str  # map to Parcels fieldset variables
+    copernicus_var: str  # map to Copernicus Marine Service variable names
+    category: Literal[
+        "phys", "bgc"
+    ]  # physical vs. biogeochemical variable, used for product ID selection logic
+    particle_vars: list[str]  # particle variable name(s) produced by this sensor
```
VeckoTheGecko: I see that this is only used in https://github.com/Parcels-code/virtualship/blob/89e184aaebf2b0128da9e06833d34442d783e72f/src/virtualship/utils.py#L746-L751 . Perhaps it would be clearer instead of these being strings for them to explicitly be lists of `parcels.Variable` objects?

Also, for my own understanding, how does this relate to (e.g.) https://github.com/Parcels-code/virtualship/blob/89e184aaebf2b0128da9e06833d34442d783e72f/src/virtualship/instruments/argo_float.py#L38-L49 ? Are the particle variables here "extra ones" needed for the sensor and to be captured as output, whereas the ones in `_ARGO_FIXED_VARIABLES` are for the base behaviour of the instrument?

src/virtualship/instruments/adcp.py
```diff
@@ -45,6 +40,11 @@ def _sample_velocity(particle, fieldset, time):
     )
 
 
+_ADCP_SENSOR_KERNELS: dict[SensorType, callable] = {
+    SensorType.VELOCITY: _sample_velocity,
+}
```
VeckoTheGecko: This type hint isn't correct
```suggestion
from collections.abc.Callable # put at top of file
_ADCP_SENSOR_KERNELS: dict[SensorType, Callable] = {
    SensorType.VELOCITY: _sample_velocity,
}
```

---

Great use of typing across the PR - I think having this typing is really useful so that it gets us to think more about the structure of the codebase. Currently I don't think typechecking is run in CI (or if it is, we currently have a lot of failures). Should we map out a path forward for enabling typechecking across the codebase as well? The types are only really useful if they're enforced (something I'm working on in Parcels as well).

src/virtualship/instruments/xbt.py
```diff
@@ -50,6 +53,11 @@ def _sample_temperature(particle, fieldset, time):
     particle.temperature = fieldset.T[time, particle.depth, particle.lat, particle.lon]
```
VeckoTheGecko, line 56: Same here about the `callable` type annotation - perhaps worth find-all'ing across the codebase for `, callable]` so all these are replaced

src/virtualship/models/expedition.py
VeckoTheGecko: To be clear, are you suggesting `from virtualship.instruments import sensors`? I think that would be a good idea :)

src/virtualship/utils.py
```diff
@@ -669,6 +761,21 @@ def _make_hash(s: str, length: int) -> str:
     return hashlib.shake_128(s.encode("utf-8")).hexdigest(half_length)
 
 
+def build_particle_class_from_sensors(
+    sensors: list[SensorConfig],
+    fixed_variables: list,
```
VeckoTheGecko:
```suggestion
    fixed_variables: list[Variable],
```

src/virtualship/utils.py
```diff
@@ -669,6 +761,21 @@ def _make_hash(s: str, length: int) -> str:
...
+    particle_class: type,
```
VeckoTheGecko: nitpick. Feel free to ignore this suggestion if you don't find it useful
```suggestion
    particle_class: type, # generic type annotation needed for v3 particle class behaviour # TODO: Update with Parcels v4
```

VeckoTheGecko (commented):
Good work so far! Really like the direction that this is going. I have a bunch of comments,  some quick fixes and others more cenceptual that I think would be good to go over together.

src/virtualship/models/expedition.py
j-atkins: Update: I think there is still a need to separate it out here because it relates to drawing these two separate fields from the Copernicus Marine Service

src/virtualship/utils.py
j-atkins: Update: also here I think this still requires separation. This maps to the particle class variable name strings which should be added as separate variables, e.g: 

```python
SampleParticle = parcels.JITParticle.add_variables(
    [
        parcels.Variable("U", dtype=np.float32, initial=np.nan),
        parcels.Variable("V", dtype=np.float32, initial=np.nan),
    ]
)
```

src/virtualship/utils.py
j-atkins: > Are the particle variables here "extra ones" needed for the sensor and to be captured as output, whereas the ones in _ARGO_FIXED_VARIABLES are for the base behaviour of the instrument?

Yes indeed that's it, and now as Erik [suggests](https://github.com/Parcels-code/virtualship/pull/328#discussion_r3071177804) this will be renamed to `*_NONSENSOR_VARIABLES`.

src/virtualship/instruments/adcp.py
j-atkins: Yes sounds good... could similar implementation as in Parcels be carried over to VirtualShip?

j-atkins:
I have addressed various comments from the first round of review on this PR. See below for a list of main changes:
- Renamed all `_*_FIXED_VARIABLES` lists to `_*_NONSENSOR_VARIABLES`
- Renamed `_SensorMeta` to `_Sensor`, added `type_: SensorType` as a field so each sensor object is self-describing, and moved the class + `SENSOR_REGISTRY` from utils.py into sensors.py
- New `_InstrumentConfigMixin` in expedition.py defines shared methods that can be flexibly inherited across the `InstrumentConfig` models, addresses duplication flagged. 
    - Note, though, also that this logic has been extended to other preexisting datetime related parts of the InstrumetConfigs, e.g. `minutes` and `lifetime` params (so this is some 'bonus' refactoring). 
    - The tests associated with each method are maintained and a new test added to test_expedition.py to check that all InstrumentConfig models inherit from the mixin.
- Instrument subclasses (e.g. `CTDInstrument`) must now define `sensor_kernel` dictionary to map the sensor types to the relevant sampling kernels as a class attribute. This is checked by the `Instrument` base class and also a test added to test_base.py. Means the allowlists for different sensor types are now explicitly defined and managed within each instrument class, which is nice for further centralising instrument characteristics and behaviours.

src/virtualship/instruments/adcp.py
VeckoTheGecko: Yes, same config etc can be used

erikvansebille (approved):

PR #333, opened 2026-05-04
Unify `CTD` and `CTD_BGC` to one instrument

### LINKED ISSUE #260: Unify `CTD` and `CTD_BGC` to one instrument

j-atkins:
Now that we have implemented the sensors refactoring in #328, the `CTD` and `CTD_BGC` instruments can nicely be consolidated to one `CTD` instrument which supports both physical and biogeochemical variable sampling (configurable via the expedition YAML).

This PR also updates the documentation to reflect these changes and updates/adds relevant tests.

Closes #260


src/virtualship/models/expedition.py
```diff
@@ -385,8 +361,7 @@ class CTD_BGCConfig(_InstrumentConfigMixin, pydantic.BaseModel):
             SensorConfig(sensor_type=SensorType.PRIMARY_PRODUCTION),
         ],
         description=(
-            "Sensors fitted to the BGC CTD. "
-            "Supported: OXYGEN, CHLOROPHYLL, NITRATE, PHOSPHATE, PH, PHYTOPLANKTON, PRIMARY_PRODUCTION. "
+            "Sensors fitted to the CTD. Supported: TEMPERATURE, SALINITY, OXYGEN, CHLOROPHYLL, NITRATE, PHOSPHATE, PH, PHYTOPLANKTON, PRIMARY_PRODUCTION. "
```
erikvansebille, line 364: Is there still a possibility to use only the physical variables? Or will a ctd now always also use bcc?

erikvansebille (commented):
Nice cleanup! But I now understand that a CTD can only be run with BGC variables; correct? So no option to use only a subset of the sensors? Would that be useful and/or authentic?

j-atkins:
> Nice cleanup! But I now understand that a CTD can only be run with BGC variables; correct? So no option to use only a subset of the sensors? Would that be useful and/or authentic?

Ah yes it is indeed possible to only use the physical variable sensors still (or any subset/combination of phys/bgc variables)! Which sensors are on/off is configurable in the instrument config section of the expedition YAML.

The docs and `virtualship plan` tool still need updating to reflect the sensor logic changes. I've made a couple of issues just now (#335, #334) to track these.

Related and maybe useful to know: this doesn't mean user can add a 'non-sensical' sensor to an instrument. E.g. you can't add a BGC variable to an ADCP. There are still controls on which sensors an instrument can take.

PR #340, opened 2026-05-19
Incomplete cache should not interfere with re-running

### LINKED ISSUE #304: Avoid incomplete cache dir interfering with subsequent runs
When running in `--difficulty-mode easy` the cache dir is not necessary. It only interferes with any subsequent runs and prompts the user to delete (which is confusing if then moving on to `medium` or `hard` when the cache should be ignored by the user).

Should probably be addressed at the point at which it throws an error (i.e. within `_unique_id` in run.py)

j-atkins:
This PR changes the cache logic so that a subsequent run is not met with an error/instruction to manually delete `cache` directory when a previous run was interrupted. This was an annoying feature previously. Now the `id_latest` will be updated to the latest timestamp because a cache from an interrupted/incomplete run is not useful.

- [x] New test added

Closes #304

### DIFF
--- a/src/virtualship/cli/_run.py
+++ b/src/virtualship/cli/_run.py
@@ -251,7 +251,7 @@ def _run(
     print(f"[TIMER] Expedition completed in {elapsed / 60.0:.2f} minutes.")
 
 
-def _unique_id(expedition: Expedition, cache_dir: Path, difficulty_level: str) -> str:
+def _unique_id(expedition: Expedition, cache_dir: Path) -> str:
     """
     Return a unique id for the expedition (marked by datetime), which can be used to determine whether the expedition has 'changed' since the last run.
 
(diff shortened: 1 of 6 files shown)


src/virtualship/cli/_run.py
```diff
@@ -255,36 +255,35 @@ def _unique_id(expedition: Expedition, cache_dir: Path) -> str:
...
-    new_id = datetime.now().strftime("%Y%m%d%H%M%S")
-    previous_id = None
+    cache_dir.mkdir(exist_ok=True)
 
-    if not cache_dir.exists():
-        cache_dir.mkdir()
-    id_path = cache_dir.joinpath(EXPEDITION_IDENTIFIER)
-    last_expedition_path = cache_dir.joinpath(EXPEDITION_LATEST)
+    id_path = cache_dir / EXPEDITION_IDENTIFIER
+    last_expedition_path = cache_dir / EXPEDITION_LATEST
```
erikvansebille: is there a reason not to use joinpath here? That is more robust to backslashes-vs-forwardslashes confusion between linux and windows? That's relevant because we're not testing on Windows (or are we?)

erikvansebille (approved):
Looks good! One small comment below

src/virtualship/cli/_run.py
j-atkins: We do test on Windows, so yes, good idea to go for more robustness, will do.

PR #342, opened 2026-05-20
Migrate to Parcels v4

### LINKED ISSUE #231: Refactor underway instruments
Underway instruments (`ADCP` and `UNDERWATER_ST`) can probably be refactored and changed from `ScipyParticle` to `JITParticle` to speed up simulations. Also avoiding looping through points (by seeding each datetime/location as separate particles?).

Check if doing so would also require a post simulation processing step to get the zarr output in the same shape as current.

Once this is implemented, it is likely that only `space_time_region` needs to be fed into `Instrument.__init__()` in [base.py](https://github.com/Parcels-code/virtualship/blob/v1-dev/src/virtualship/instruments/base.py), as discussed [here](https://github.com/Parcels-code/virtualship/pull/226#discussion_r2527690321).

### LINKED ISSUE #232: Integrate Parcels v4 into VirtualShip
Integrate Parcels v4 into VirtualShip, can be an experimental branch for now and merged/released when Parcels v4 itself is released.

This should see a lot of code evaporate, e.g. `space_time_region`, and ease some existing bottlenecks.

### LINKED ISSUE #238: Add support for Python 3.13/3.14
[pyproject.toml](https://github.com/Parcels-code/virtualship/blob/main/pyproject.toml) points up to 3.12

### LINKED ISSUE #239: Update numpy dependency
Currently limited to < v2

### LINKED ISSUE #323: .csv output
It would be good to add (back) .csv output for CTDs, and potentially other instruments. This would better represent real-life output and potentially mean VirtualShip output could be used by people with less technical backgrounds.

j-atkins:
This PR migrates to using Parcels v4 for VirtualShip. 

----

- [x] Update instrument kernels
- [x] Update fieldset ingestion workflows
- [x] Update to Parquet output
- [x] Implement new fieldset/performance handling strategy, i.e. new [v4 docs](https://github.com/erikvansebille/Parcels/blob/1395965ef300ad86022d382349eace687b8f171b/docs/user_guide/examples/explanation_performance.md)
- [x] Implement #231 
- [x] Check expedition / workflow
    - [x] including underway instruments
    - [x] ~~Argo float bug [in kernel(s)?], immediately goes out of bounds after first timestep... likely depth convention related~~
- [x] Implement initial timestep sampling (i.e. Parcels [#2622](https://github.com/Parcels-code/Parcels/pull/2622))
- [x] Update to work with local data / `--from-data`
- [x] Update tests
- [x] Add new tests for v4 specific logic(?)
- [x] Update docs / tutorials (e.g. notebooks using `.zarr` output) [new PR]
    - [x] Plus new requirements for local data ingestion: depth dim has `positive: down` or `positive: up`
- [x] `ARGO_FLOATS`: Enhance performance + no need to update `cycle_age` during ascent?

----

Closes #239, closes #232, closes #231, closes #323, closes #238


j-atkins:
Hi @erikvansebille and @VeckoTheGecko, here's the latest draft of VirtualShip migration to Parcels v4 (still a WIP though). No need to review properly at this point, it's for your reference in case you are curious whilst I'm away over the next couple of weeks!

j-atkins:
cc @erikvansebille 

This PR is now ready for review. I need to update the docs (including the post-processing tutorials to handle the new parquet data), and I will do in a separate PR, but there's enough here to go through in the meantime. Most of it has already been reviewed in previous PRs.

j-atkins:
cc @erikvansebille 

This PR is ready for re-review. Please note a couple of things, which have been added more recently, including tinkering around `_via_tmp_ds` in base.py (e.g. using Zarr stores to try and squeeze out a bit more performance), plus moving to execute instrument simulations with a context manager protocol for a better handling of the tmp directory writing. Also, changing the Argo Float `DT` (now matching the 5 min output dt) which speeds up the actual Parcels simulation considerably (unless there is a reason to keep it as `DT = 10.0`?). 

Also, there are some changes that have been reviewed already, most notably the updates to the docs, and probably don’t need much time spent on them. Apologies this probably wasn’t the most efficient way of structuring the PRs.

src/virtualship/instruments/base.py
```diff
@@ -97,51 +135,117 @@ def load_input_data(self) -> FieldSet:
...
+            # non-underway instruments to windowed arrays, just in case any ds is Dask backed
+            # underway instruments should not to converted to windowed arrays, as they use one direct fieldset.eval() call which could cause a big memory usage if the fieldset is windowed
+            if not self.instrument_type.is_underway:
+                fs = fs.to_windowed_arrays()
+
+            fieldsets_list.append(fs)
+
+        combined_fieldset = fieldsets_list[0]
+        for fs in fieldsets_list[1:]:
+            combined_fieldset = combined_fieldset + fs
```
erikvansebille: Can you not simply keep adding the fieldsets? Instead of first making a list and then adding the entries of that list?

src/virtualship/instruments/ctd.py
```diff
@@ -52,60 +53,69 @@ class CTD:
...
+def _sample_temperature(particles, fieldset):
+    particles.temperature = fieldset.T[
+        particles.t, particles.z, particles.y, particles.x
+    ]
 
 
-def _sample_salinity(particle, fieldset, time):
-    particle.salinity = fieldset.S[time, particle.depth, particle.lat, particle.lon]
+def _sample_salinity(particles, fieldset):
+    particles.salinity = fieldset.S[particles.t, particles.z, particles.y, particles.x]
```
erikvansebille: In general, it's better to sample with `[particles`] since that retains the `particles._ei` element indices, so makes grid searching faster

```suggestion
    particles.salinity = fieldset.S[particles]
```

Also relevant for the other sampling kernels

src/virtualship/instruments/ctd.py
```diff
@@ -52,60 +53,69 @@ class CTD:
...
-        particle_ddepth = -particle.winch_speed * particle.dt
-        if particle.depth + particle_ddepth < particle.max_depth:
-            particle.raising = 1
-            particle_ddepth = -particle_ddepth
+    particles_lowering.dz += -particles_lowering.winch_speed * particles_lowering.dt
+    particles_lowering.raising = np.where(
+        particles_lowering.z + particles_lowering.dz < particles_lowering.max_depth,
+        1,
+        particles_lowering.raising,
+    )
```
erikvansebille: Not sure how you feel, but I tend to prefer boolean indexing (like in the Argo float kernel) over np.where(), as I find the boolean indexing easier to read

src/virtualship/instruments/ctd.py
```diff
@@ -208,24 +192,27 @@ def simulate(self, measurements, out_path) -> None:
...
             fieldset=fieldset,
             pclass=_CTDParticle,
-            lon=[ctd.spacetime.location.lon for ctd in measurements],
-            lat=[ctd.spacetime.location.lat for ctd in measurements],
-            depth=[ctd.min_depth for ctd in measurements],
-            time=[ctd.spacetime.time for ctd in measurements],
+            x=[ctd.spacetime.location.lon for ctd in measurements],
+            y=[ctd.spacetime.location.lat for ctd in measurements],
+            z=[ctd.min_depth for ctd in measurements],
+            t=[np.datetime64(ctd.spacetime.time) for ctd in measurements],
```
erikvansebille: why si this casting to datetime64 needed? Can't it be done without?

src/virtualship/models/expedition.py
```diff
@@ -147,7 +139,7 @@ def verify(
             for wp_i, wp in enumerate(self.waypoints):
                 try:
                     value = bathymetry_field.eval(
-                        0,  # time
+                        np.float64(0.0),  # time
```
erikvansebille: Interesting that you needed this change. Is this a bug in Parcels?

pixi.toml
erikvansebille, line 1: perhaps let @VeckoTheGecko also double-check the pinnings in this `pixi.toml`?

erikvansebille (commented):
Impressive effort, @j-atkins! See below some small comments/questions

src/virtualship/instruments/ctd.py
j-atkins: I get an error when using the raw `datetime.datetime` objects:

```
RuntimeError: An unexpected error occurred while simulating measurements: particle t must be a datetime, timedelta, or date object.
```

And I find the same error outside of VirtualShip. Given the error messaging suggests `datetime` objects should be supported, I presume this is a bug. I'll can make an issue on the Parcels tracker!

src/virtualship/models/expedition.py
j-atkins: I think this is leftover from a workaround for a bug which may have existed previously (?) but is now seemingly resolved in Parcels. Therefore, will be removed.

j-atkins:
Thanks for the review, @erikvansebille! Comments now addressed, including moving to using boolean indexing across all (non-underway) instrument kernels.

@VeckoTheGecko I have updated/unpinned pixi in the `pixi.toml` so should be up to date. Do you just want to have a quick once over though?

VeckoTheGecko:
> I have updated/unpinned pixi in the `pixi.toml` so should be up to date. Do you just want to have a quick once over though?

Heya - just seeing this. Will take a look at the `pixi.toml`

VeckoTheGecko:
The `pixi.toml` looks good! (only question is about `sortedcontainers`, which is a dependency but not used anywhere in the project - but that's not from this PR)

erikvansebille (approved):
Looks good - happy for you to merge!

PR #348, opened 2026-05-29
`FieldOutOfBoundError` does not exit simulation

### LINKED ISSUE #154: Capture FieldOutOfBound errors on drifters and Argo floats
When a drifter (or Argo) leaves the downloaded domain, this now throws a `FieldOutOfBounds` error that will stop the entire `virtualship run`. It would be better to try/except these errors, throw a warning and then at least continue.

j-atkins:
Following on from #154 and discussion therein, this PR addresses `FieldOutOfBoundErrors` for Argo Floats. 

Previously errors were captured via the `_check_error()` kernel but this silently deleted particles (see https://github.com/Parcels-code/virtualship/pull/33#discussion_r1616087895). Now, the kernel will print a warning so users at least know to expect incomplete data and can consider fixes.

- [x] New test added which checks simulation has continued despite out of bounds error (also refactored/reduced duplication in test_argo_float.py)

Closes #154

### DIFF
--- a/src/virtualship/instruments/argo_float.py
+++ b/src/virtualship/instruments/argo_float.py
@@ -131,6 +131,23 @@ def _keep_at_surface(particle, fieldset, time):
 
 def _check_error(particle, fieldset, time):
     if particle.state >= 50:  # This captures all Errors
+        if particle.state == 50:
+            print("WARNING: Error during Argo Float simulation...")
+        elif particle.state == 51:
+            print("WARNING: ErrorInterpolation during Argo Float simulation...")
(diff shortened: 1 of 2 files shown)


src/virtualship/instruments/argo_float.py
erikvansebille, line 1: In v4 you may indeed want to do this more pythonic with dictionaries (`{50: "Error", 51: "ErroInterpolation", ...}`)?

erikvansebille (approved):
Looks good!

Issue #357, opened 2026-07-16
VirtualShip v1 / Parcels v4 performance


j-atkins:
## Summary

This issue is to acknowledge and track the progress of Parcels v4 performance with `copernicusmarine` (see [#2756](https://github.com/Parcels-code/Parcels/issues/2756)) and its relevance to VirtualShip v1.

## More detail

A core part of an upcoming VirtualShip v1 release (#195) will mean the move from Parcels v3 -> v4 (#342) under the hood. In line with the aims of the Parcels v4 release, this overhauls the VirtualShip internals to align more closely with the Pangeo ecosystem and use vectorised operations.

Parcels v4 computation works and is ready for use in VirtualShip simulations, but its performance will continue to be improved as v4 evolves. Latest performance testing suggests that local data ingestion is performant but streaming workflows (e.g. with `copernicusmarine`) are slower. This is relevant to VirtualShip, where a key feature is the default Digital Twin/streaming nature of the framework. Thus, we should expect to see some performance decrease on initial VirtualShip v1 release. However, ongoing work on the Parcels side to increase performance in v4.x... will feed directly into VirtualShip.

j-atkins:
The next step for VirtualShip v1 development will be to adopt the tmp file writing strategy as demonstrated in the [Parcels benchmarking](https://github.com/Parcels-code/parcels-benchmarks/pull/49).

j-atkins:
The tmp file writing strategy has been adopted in #358.

Note though that the writing to disk step in the case of Argo Floats is very slow (high-volume 4d data). It would be good to explore speeding this up.

PR #358, opened 2026-07-17
add tmp file write/read step

j-atkins:
This PR implements a write-to-tmp files step, to improve performance for Parcels v4 fieldset ingestion combined with copernicusmarine streaming (i.e. #357).

**UPDATE**: As posted in comment below:
>This is now ready for review and appears the most performant strategy for VirtualShip for now.
>
>- [x] New test for `_via_tmp_ds()`

Please note test failures are expected, and will be addressed in root #342 PR

### DIFF
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -239,6 +239,7 @@ class Instrument(abc.ABC):
             ds_fset = self._via_tmp_ds(ds_fset)
 
             fs = parcels.FieldSet.from_sgrid_conventions(ds_fset)
+            fs.to_windowed_arrays()  # always to windowed arrays, just in case any ds is Dask backed
 
             fieldsets_list.append(fs)
 
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import abc
 import collections
+import tempfile
 from datetime import timedelta
 from itertools import pairwise
 from pathlib import Path
@@ -183,6 +184,9 @@ class Instrument(abc.ABC):
         Create and combine FieldSets for each variable, supporting both local and Copernicus Marine data sources.
 
         N.B. Per variable avoids issues when using copernicusmarine and creating directly one FieldSet of ds's sourced from different Copernicus Marine product IDs (which can also have different temporal resolutions), which is often the case for BGC variables.
+
+        Includes an intermediate step of writing to tmp files, as per https://github.com/Parcels-code/parcels-benchmarks/pull/49
+        TODO: the need for this step may be removed as Parcels x copernicusmarine integration improves, tracked in https://github.com/Parcels-code/Parcels/issues/2756 and xref'd in VirtualShip #357 (https://github.com/Parcels-code/virtualship/issues/357)
         """
         fieldsets_list = []
         keys = list(self.variables.keys())
@@ -222,11 +226,12 @@ class Instrument(abc.ABC):
             ds["depth"] = -ds["depth"]
             ds = ds.reindex(depth=ds["depth"][::-1])
 
-            # TODO: update when decision on handling of nans/0s in v4 is made (i.e. https://github.com/Parcels-code/Parcels/issues/2393)
+            # TODO: to be removed when Parcels #2746 is merged (i.e. https://github.com/Parcels-code/Parcels/pull/2746)
             ds = ds.fillna(0)
 
             fields = {key: ds[field_var_name]}
             ds_fset = parcels.convert.copernicusmarine_to_sgrid(fields=fields)
+            ds_fset = self._via_tmp_ds(ds_fset)
 
             fs = parcels.FieldSet.from_sgrid_conventions(ds_fset)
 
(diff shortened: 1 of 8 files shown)


j-atkins:
@erikvansebille, initial experimentation shows that the tmp file performance fix is indeed a simple implementation and substantially improves actual computation performance, but the tradeoff here is a slow-down by the fact that VirtualShip, in most cases, calls to global datasets from `copernicusmarine`. 

In the case of e.g. a Drifter simulation the `ds`'s (single level) are therefore ~6GB and are not super quick to write/read to/from disk, resulting in a simple single Drifter simulation (63 day lifetime) taking about 3.5 minutes all in, the majority of which time is writing the file to disk. Should note though that this is still a performance increase relative to `.to_windowed_array()` attempts!

A `.sel()` to a smaller spatial window (see quick example below) doesn't improve much, presumably something to do with the comms across the streaming.

```python
  def _via_tmp_ds(self, ds) -> xr.Dataset:
      """Create and re-load a temporary local dataset."""
      tmpdir = tempfile.TemporaryDirectory()
      tmp_fpath = Path(tmpdir.name).joinpath("tmp.nc")

      # reduce dataset to smaller spatial window to reduce filesize
      HALO_DEGREES = 20.0  # TODO ...arbitrary but want a healthy buffer for drifters
      ds_minimised = ds.sel(
          lat=slice(self.min_lat - HALO_DEGREES, self.max_lat + HALO_DEGREES),
          lon=slice(self.min_lon - HALO_DEGREES, self.max_lon + HALO_DEGREES),
      )

      ds_minimised.to_netcdf(tmp_fpath)
      del ds, ds_minimised

      return xr.open_dataset(tmp_fpath)
```

I'll experiment further with file size reduction strategies (e.g. at the `copernicusmarine.open_dataset()` stage) to improve speed.

Of course the motivation for calling to global datasets in the first place was to avoid any risk of out of bounds errors so that will also be a priority.

j-atkins:
>
>In the case of e.g. a Drifter simulation the ds's (single level) are therefore ~6GB and are not super quick to write/read >to/from disk, resulting in a simple single Drifter simulation (63 day lifetime) taking about 3.5 minutes all in, the majority of >which time is writing the file to disk. Should note though that this is still a performance increase relative to `.to_windowed_array()` attempts!
>

Following up on this, the file writing speeds don't appear to be overwhelmingly tied to the the spatial extent of the requested ds, in fact they appear more sensitive to the temporal extent (perhaps to do with chunking strategy somewhere on the `copernicusmarine` end).

Performance for other instruments, such as CTDs and Argo Floats, is pretty good relative to using Parcels v3 though. So, overall, it's promising and it appears that it's just Drifters which slow down (a bit but not massively) relative to v3.

j-atkins:
I think it would probably be a good idea to reduce the file sizes for Drifter simulations, even if just tmp files, because ~6GB global datasets feel a bit unnecessary and potentially problematic. Even adding a generous spatial buffer (+/- 30 degrees) would reduce the size substantially whilst avoiding out of bounds errors in nearly all cases. 

This would effectively create a hard limit on lifetimes (velocity dependent) but perhaps a necessary (temporary) trade-off for now. Also, at least this wouldn't crash the simulations, as of #348...

What do you think @erikvansebille?

erikvansebille:
> I think it would probably be a good idea to reduce the file sizes for Drifter simulations, even if just tmp files, because ~6GB global datasets feel a bit unnecessary and potentially problematic. Even adding a generous spatial buffer (+/- 30 degrees) would reduce the size substantially whilst avoiding out of bounds errors in nearly all cases.
> 
> This would effectively create a hard limit on lifetimes (velocity dependent) but perhaps a necessary (temporary) trade-off for now. Also, at least this wouldn't crash the simulations, as of #348...
> 
> What do you think @erikvansebille?

Yes, I agree that using a buffer for the drifters is the most effective solution for now. 

Hopefully, as we improve the streaming performance in Parcels v4 (see https://github.com/Parcels-code/parcels-benchmarks/pull/49), we can then remove the buffer again. But I have no idea if that performance improvement will take one week or half a year...

j-atkins:
This is now ready for review and appears the most performant strategy for VirtualShip for now.

- [x] New test for `_via_tmp_ds()`

src/virtualship/instruments/base.py
```diff
@@ -257,3 +258,12 @@ def _generate_fieldset(self) -> parcels.FieldSet:
...
         return base_fieldset
+
+    @staticmethod
+    def _via_tmp_ds(ds) -> xr.Dataset:
+        """Create and re-load a temporary local dataset."""
+        tmpdir = tempfile.TemporaryDirectory()
+        tmp_fpath = Path(tmpdir.name).joinpath("tmp.nc")
+        ds.to_netcdf(tmp_fpath)
+        del ds
+        return xr.open_dataset(tmp_fpath)
```
erikvansebille, line 270: Is it worth checking/making sure what the backend is, here? Either Dask or NumPy? See also the (draft) performance guide at https://parcels--2705.org.readthedocs.build/en/2705/user_guide/examples/explanation_performance.html. 

If it turns out to use Dask under the hood, then make sure to also do a `to_windowed_arrays` call; otherwise performance could still be poor

erikvansebille (commented):
Looks good, just one comment below

src/virtualship/instruments/base.py
j-atkins: The fieldsets are consistently NumPy backed across different instruments and checked against large/long expeditions. As discussed though, I have added `to_windowed_arrays` as a safe option, just in case there are rare scenarios where Dask is used.

erikvansebille (approved):

PR #359, opened 2026-07-21
Refactor dataset retrieval specs

j-atkins:
This PR is a small refactoring for the migrate-v4 branch. The previous `limit_spec` and `spacetime_buffer_size` logic was getting messy and unclear how the two dicts should interact. I find adding a `FetchSpec` dataclass in base.py to be clearer and more explicit about the requirements for `Instrument` classes and clearer about what the defaults are.

### DIFF
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -50,8 +50,7 @@ class Instrument(abc.ABC):
         allow_time_extrapolation: bool,
         verbose_progress: bool,
         from_data: Path | None,
-        spacetime_buffer_size: dict | None = None,
-        limit_spec: dict | None = None,
+        fetch_spec: dict | None = None,
     ):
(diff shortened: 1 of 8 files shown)


src/virtualship/instruments/base.py
```diff
@@ -28,6 +29,17 @@
     from virtualship.models import Expedition
 
 
+@dataclass
+class FetchSpec:
```
erikvansebille, line 33: Does this class need its own unit tests? Or is it tested sufficiently with the instrument tests?

erikvansebille (approved):
Looks good; just one question below

src/virtualship/instruments/base.py
j-atkins: Added an additional test for thoroughness 👍

erikvansebille (approved):

PR #360, opened 2026-07-23
Update underway

### LINKED ISSUE #231: Refactor underway instruments

j-atkins:
This PR implements a move to using `Field.eval` for underway instrument simulations (ADCP, UnderwaterST). This is enabled by Parcels v4 and delivers much faster performance.

To facilitate this, there is a new abstract/intermediate class for underway instruments in base.py: `UnderwayInstrument`, which the ADCP and UnderwaterST concrete subclasses inherit from. Plus a new `UnderwayCoordinates` dataclass for housing the final sampling locations and timings. There are new tests for these in test_base.py

Also, underway instruments still have 'kernels' but are not traditional Parcels kernels. Instead, they are standard functions which make the `Field.eval` call. This means underway instruments continue to meet the requirements for all instruments to have sampling kernels associated with sensors.

---

Closes #231

### DIFF
--- a/src/virtualship/instruments/adcp.py
+++ b/src/virtualship/instruments/adcp.py
@@ -96,6 +96,7 @@ class ADCPInstrument(Instrument):
                 for point in measurements
             ]
         )
+
         lons = np.array([point.location.lon for point in measurements])
         lats = np.array([point.location.lat for point in measurements])
         bins = np.linspace(MAX_DEPTH, MIN_DEPTH, NUM_BINS)
@@ -105,7 +106,10 @@ class ADCPInstrument(Instrument):
         lats_full = np.repeat(lats, NUM_BINS)
         depths_full = np.tile(bins, len(times))
 
-        u, v = fieldset.UV.eval(t=times_full, z=depths_full, x=lons_full, y=lats_full)
+        u, v = (
+            fieldset.U.eval(t=times_full, z=depths_full, x=lons_full, y=lats_full),
+            fieldset.V.eval(t=times_full, z=depths_full, x=lons_full, y=lats_full),
+        )
 
         _write_underway_to_parquet(
             dat_arrays=[u, v],
--- a/src/virtualship/instruments/adcp.py
+++ b/src/virtualship/instruments/adcp.py
@@ -113,7 +113,7 @@ class ADCPInstrument(UnderwayInstrument):
             config_sensors=adcp_config.sensors, fieldset=fieldset, coords=coords
         )
 
-        self._write_underway_to_parquet(
+        self._to_parquet(
             dat_arrays=sampled,
             var_names=self.variables.keys(),
             fieldset_time_origin=fieldset_starttime,
--- a/src/virtualship/instruments/adcp.py
+++ b/src/virtualship/instruments/adcp.py
@@ -1,38 +1,35 @@
 from collections.abc import Callable
-from dataclasses import dataclass
 from typing import ClassVar
 
 import numpy as np
 
-from virtualship.instruments.base import FetchSpec, Instrument
+from virtualship.instruments.base import (
+    FetchSpec,
+    UnderwayCoordinates,
+    UnderwayInstrument,
+)
 from virtualship.instruments.sensors import SensorType
 from virtualship.instruments.types import InstrumentType
-from virtualship.utils import _write_underway_to_parquet, register_instrument
-
-# =====================================================
-# SECTION: Dataclass
-# =====================================================
-
-
-@dataclass
-class ADCP:
-    """ADCP configuration."""
-
-    name: ClassVar[str] = "ADCP"
-
+from virtualship.utils import register_instrument
 
 # =====================================================
 # SECTION: Kernels
 # =====================================================
 
 
-def _sample_velocity(particles, fieldset):
-    particles.U, particles.V = fieldset.UV.eval(
-        particles.t,
-        particles.z,
-        particles.x,
-        particles.y,
-    )
+# N.B. underway 'kernels' are special cases, where the particleset is not needed, and the kernel is not passed to `pset.execute()` as would be done for a typical Parcels workflow.
+# Instead, the 'kernel' function is used only once to evaluate the fieldset at given times, depths, lats, lons.
+
+
+def _sample_underway_velocity(fieldset, times_full, depths_full, lats_full, lons_full):
+    # eval
+    u, v = fieldset.UV.eval(t=times_full, z=depths_full, x=lons_full, y=lats_full)
+
+    # convert from degrees s-1 to metres s-1
+    u = u * 1852 * 60 * np.cos(np.deg2rad(lats_full))
+    v = v * 1852 * 60
+
+    return u, v
 
 
 # =====================================================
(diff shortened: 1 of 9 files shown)


src/virtualship/instruments/adcp.py
```diff
@@ -87,53 +77,46 @@ def simulate(self, measurements, out_path) -> None:
...
         fieldset = self.load_input_data()
 
-        # build dynamic particle class from the active sensors
-        adcp_config = self.expedition.instruments_config.adcp_config
-        _ADCPParticle = build_particle_class_from_sensors(
-            adcp_config.sensors, _ADCP_NONSENSOR_VARIABLES
+        # use first active field for time reference
+        _time_ref_key = next(iter(self.variables))
+        _time_ref_field = getattr(fieldset, _time_ref_key)
+        fieldset_starttime = _time_ref_field.data.time.isel(time=0).values
```
erikvansebille: Could you also use `fieldset.time_interval.left` here?

src/virtualship/instruments/base.py
```diff
@@ -279,3 +286,143 @@ def _via_tmp_ds(ds) -> xr.Dataset:
     def instrument_type(self) -> InstrumentType:
         """Return the InstrumentType for this instrument instance."""
         return next(k for k, v in INSTRUMENT_CLASS_MAP.items() if type(self) is v)
+
+
+@dataclass(frozen=True)
+class UnderwayCoordinates:
+    """1D evaluation grid arrays for underway instruments."""
```
erikvansebille: what do you mean with "grid arrays" here?

src/virtualship/instruments/base.py
```diff
@@ -279,3 +286,143 @@ def _via_tmp_ds(ds) -> xr.Dataset:
...
+        table = pa.table(
+            {
+                "t": pa.array(coords.times.astype(np.float64)),
+                "z": pa.array(coords.depths.astype(np.float32))
+                if coords.depths is not None
+                else pa.array(np.full(n, np.nan, dtype=np.float32)),
+                "y": pa.array(coords.lats.astype(np.float32)),
+                "x": pa.array(coords.lons.astype(np.float32)),
+                "particle_id": pa.array(
+                    np.arange(n, dtype=np.int64)
```
erikvansebille: Would it not make sense to give all data the same particle ID (interpreting the ship as one particle)?

src/virtualship/instruments/base.py
```diff
@@ -279,3 +286,143 @@ def _via_tmp_ds(ds) -> xr.Dataset:
...
+        assert len(dat_arrays) == len(var_names), (
+            "dat_arrays and var_names must have the same length"
+        )
+
+        n = len(coords.times)
+
+        origin_str = str(fieldset_time_origin).replace("T", " ")
+        t_metadata = {"units": f"seconds since {origin_str}", "calendar": "standard"}
+
+        # base schema mirroring Parcels ParticleFile schema, not yet with sampled variables
```
erikvansebille, line 372: Would it make sense to also open a parcels.ParticleFile object to check if the schemes are the same - and error out if they are not? In that way, we can spot when the schemas start to deviate

src/virtualship/instruments/ship_underwater_st.py
```diff
@@ -80,49 +69,41 @@ def __init__(self, expedition, from_data):
...
         fieldset = self.load_input_data()
 
-        # build dynamic particle class from the active sensors
-        st_config = self.expedition.instruments_config.ship_underwater_st_config
-        _ShipSTParticle = build_particle_class_from_sensors(
-            st_config.sensors, _ST_NONSENSOR_VARIABLES
+        # use first active field for time reference
+        _time_ref_key = next(iter(self.variables))
+        _time_ref_field = getattr(fieldset, _time_ref_key)
+        fieldset_starttime = _time_ref_field.data.time.isel(time=0).values
```
erikvansebille: Same comment as with ADCP; use `fieldset.time_interval.left?

erikvansebille (commented):
Looks good; a few questions/suggestions below

src/virtualship/instruments/base.py
j-atkins: Yes, perhaps this is confusing, will update

src/virtualship/instruments/base.py
j-atkins: I have now added a new unit test for this (in https://github.com/Parcels-code/virtualship/pull/360/commits/ccd0ae3c17e847cc43eb9fcc282da5d5fd3c4474). The scheme is not in the public API for `parcels.PartileFile` so I have done it by executing a simple Parcels simulation and comparing the scheme of the resultant output against the `UnderwayInstrument` output.

tests/instruments/test_base.py
```diff
@@ -193,3 +208,245 @@ def test_instrument_subclass_without_sensor_kernels_error():
...
+        out_path=out_path,
+        coords=coords,
+    )
+
+
+def dummy_sample_temperature(particles, fieldset):
+    particles.T = fieldset.T[particles.t, particles.z, particles.y, particles.x]
+
+
+def test_parquet_openable_by_read_particles(tmp_path):
```
erikvansebille:
```suggestion
def test_parquet_openable_by_parcels_read_particlefile(tmp_path):
```

j-atkins:
Thanks for the comments @erikvansebille, I have now made the changes in response 👍

tests/instruments/test_base.py
```diff
@@ -193,3 +208,245 @@ def test_instrument_subclass_without_sensor_kernels_error():
...
+
+    pset = parcels.ParticleSet(
+        fieldset=fieldset, pclass=SampleParticle, t=t1, y=[0.0], x=[0.0]
+    )
+
+    parcels_path = tmp_path / "parcels_particles.parquet"
+    parcels_output = parcels.ParticleFile(parcels_path, outputdt=3600.0)
+    pset.execute(
+        [dummy_sample_temperature],
+        runtime=np.timedelta64(1, "D"),
```
erikvansebille: If the goal is to only check the schema, then why do you need to run for so long? Would one write not be sufficient?

src/virtualship/instruments/base.py
```diff
@@ -132,7 +139,7 @@ def simulate(
 
     def execute(self, measurements: list, out_path: str | Path) -> None:
         """Run instrument simulation."""
-        TMP = True  # TODO: just for dev; remove before merging
+        TMP = False  # TODO: just for dev; remove before merging
```
erikvansebille: Should this be removed then?

erikvansebille (approved):
Three small points/questions; looks good otherwise!

tests/instruments/test_base.py
j-atkins: Yes, good point, I had just defaulted to using something simple like this but it's indeed unnecessary

PR #361, opened 2026-07-27
`to_windowed_arrays` only for non-underway instruments

j-atkins:
This PR adds a check if an instrument is/not an underway instrument (`ADCP`, `UnderwaterST`). If not, then use `to_windowed_arrays`. Otherwise leave as the default NumPy/Dask backed fieldset the `fieldset.eval` in the underway simulation methods.

All instruments now all have a property method for their `InstrumentType` (in base.py), which means the `is_underway` attribute becomes accessible. Related unit tests added for all instruments.

Please note test failures are expected, and will be addressed in root https://github.com/Parcels-code/virtualship/pull/342 PR

### DIFF
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -243,9 +243,9 @@ class Instrument(abc.ABC):
             fs = parcels.FieldSet.from_sgrid_conventions(ds_fset)
 
             # non-underway instruments to windowed arrays, just in case any ds is Dask backed
-            # underway instruments not to windowed arrays, they use fieldset.eval() which could cause a big memory usage if the fieldset is windowed
+            # underway instruments should not to converted to windowed arrays, as they use one direct fieldset.eval() call which could cause a big memory usage if the fieldset is windowed
             if not self.instrument_type.is_underway:
-                fs.to_windowed_arrays()
+                fs = fs.to_windowed_arrays()
 
             fieldsets_list.append(fs)
 
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -132,7 +132,7 @@ class Instrument(abc.ABC):
 
     def execute(self, measurements: list, out_path: str | Path) -> None:
         """Run instrument simulation."""
-        TMP = False  # TODO: just for dev; remove before merging
+        TMP = True  # TODO: just for dev; remove before merging
         instrument_name = self.__class__.__name__.split("Instrument")[0]
 
         if not self.verbose_progress:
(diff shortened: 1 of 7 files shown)


src/virtualship/instruments/base.py
```diff
@@ -239,7 +241,11 @@ def _generate_fieldset(self) -> parcels.FieldSet:
             ds_fset = self._via_tmp_ds(ds_fset)
 
             fs = parcels.FieldSet.from_sgrid_conventions(ds_fset)
-            fs.to_windowed_arrays()  # always to windowed arrays, just in case any ds is Dask backed
+
+            # non-underway instruments to windowed arrays, just in case any ds is Dask backed
+            # underway instruments not to windowed arrays, they use fieldset.eval() which could cause a big memory usage if the fieldset is windowed
+            if not self.instrument_type.is_underway:
+                fs.to_windowed_arrays()
```
erikvansebille: Not entirely sure if to_windowed_arrays() works inlace; safer to do the following?
```suggestion
                fs = fs.to_windowed_arrays()
```

src/virtualship/instruments/base.py
```diff
@@ -239,7 +241,11 @@ def _generate_fieldset(self) -> parcels.FieldSet:
             ds_fset = self._via_tmp_ds(ds_fset)
 
             fs = parcels.FieldSet.from_sgrid_conventions(ds_fset)
-            fs.to_windowed_arrays()  # always to windowed arrays, just in case any ds is Dask backed
+
+            # non-underway instruments to windowed arrays, just in case any ds is Dask backed
+            # underway instruments not to windowed arrays, they use fieldset.eval() which could cause a big memory usage if the fieldset is windowed
```
erikvansebille:
```suggestion
            # underway instruments should not to converted to windowed arrays, as they use one direct fieldset.eval() call which could cause a big memory usage if the fieldset is windowed
```

tests/instruments/test_adcp.py
erikvansebille, line 1: Are all these changed test files also part of this PR? Should I review these too?

j-atkins: Yes they are part of this PR. It is the same across each instrument though: adding a test which confirms that the `.instrument_type` property works and returns the correct `InstrumentType`, plus whether it's an underway instrument or not. 

I considered having this kind of check in one place in test_base.py but this would likely mean explicitly importing all separate instrument types and therefore add another file to be changed when adding instruments. Whereas, adding a new `test_{instrument_name}.py` file would already be necessary with new instruments.

The test changes also contain refactorings, but only to avoid repeatedly making the same (or similar) `DummyExpedition` classes.

erikvansebille (approved):
Looks good. Happy to be merged when green ticks

j-atkins:
> Looks good. Happy to be merged when green ticks

Great, thanks. Note, certain unit tests are expected to fail (meaning we won't get all green ticks in the PR) but will be addressed in the target #342 branch/PR. They are left to fail for now though to ensure they do not get missed later.

PR #364, opened 2026-07-31
Add initial Field sampling for non-underway instruments

j-atkins:
This PR implements initial Field sampling for non-underway instruments (i.e. [Parcels #2622](https://github.com/Parcels-code/Parcels/pull/2622)), via a new `_sample_initial` method in the `Instrument` base class. 

- [x] New unit test for `_sample_initial`

### DIFF
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -297,7 +297,7 @@ class Instrument(abc.ABC):
             particle_vars = [pv.name for pv in sensor.meta.particle_vars]
 
             for var in particle_vars:
-                setattr(pset, var, field[pset.t, pset.z, pset.y, pset.x])
+                setattr(pset, var, field[pset])
 
         return pset
 
--- a/src/virtualship/instruments/base.py
+++ b/src/virtualship/instruments/base.py
@@ -155,6 +155,26 @@ class Instrument(abc.ABC):
             self.simulate(measurements, out_path)
             print("\n")
 
+    def _sample_initial(
+        self,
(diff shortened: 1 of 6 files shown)


src/virtualship/instruments/base.py
```diff
@@ -279,6 +279,28 @@ def _via_tmp_ds(ds) -> xr.Dataset:
...
+                raise ValueError(
+                    f"Attempted to initialise sensor '{sensor.sensor_type}' but it is not enabled in the expedition configuration."
+                )
+
+            fs_key = sensor.meta.fs_key
+            field = getattr(fieldset, fs_key)
+            particle_vars = [pv.name for pv in sensor.meta.particle_vars]
+
+            for var in particle_vars:
+                setattr(pset, var, field[pset.t, pset.z, pset.y, pset.x])
```
erikvansebille: Better to use `pset` as an argument here, which makes sure that the element indices (`pset._ei`) also get updated
```suggestion
                setattr(pset, var, field[pset])
```
Note that this only works with the recently merged https://github.com/Parcels-code/Parcels/pull/2622

erikvansebille (commented):
Looks good, just one suggestion below

src/virtualship/instruments/base.py
j-atkins: Ah right I see, this makes sense now. It didn't work previously with just `pset` as the argument but I hadn't accounted for the (un-merged at the time) changes in Parcels #2622. Thanks!

erikvansebille (approved):

PR #371, opened 2026-08-11
Update docs for v4 migration

j-atkins:
This PR updates the docs, including tutorials and technical documentation, to match any changes associated with the migration to Parcels v4 (#342).


j-atkins:
cc @erikvansebille 

Please note, full commit history is missing from this PR (apologies if this makes reviewing cumbersome). I was having buggy/phantom diffs relative to the migrate-v4 branch, so made a clean new branch with changes transferred over in one swoop.

docs/user-guide/documentation/pre_download_data.md
```diff
@@ -77,6 +77,7 @@ The following assumptions are also made about the data:
 3. If using BGC-enabled instruments (e.g. BGC variables on the `CTD`), the relevant biogeochemical data files must contain the following variables: `o2`, `chl`, `no3`, `po4`, `nppv`, `ph`, `phyc`.
    - Or these strings must appear as substrings within the variable names (e.g. `o2_glor` is acceptable for `o2`).
 4. Bathymetry data files must contain a variable named `deptho`.
+5. Pre-downloaded data files must have a `"positive"` attribute for the depth dimension (e.g. `"positive": "down"` or `"positive": "up"`) in order to ensure that the depth dimension is correctly interpreted by Parcels under-the-hood.
```
erikvansebille: Is this a Parcels requirement? Or a VirtualShip requirement?

erikvansebille (approved):
Looks good! One comment below

docs/user-guide/documentation/pre_download_data.md
j-atkins: This is a (new) VirtualShip requirement. Previously the step to make the depth dimension negative and positive: up was generic across both streaming and local ingestion pipelines. Now that we use `vertical_axis="elevation"` in the `copernicusmarine.open_dataset()` [call](https://github.com/Parcels-code/virtualship/blob/26d6681b761722d436ab82e0babc97f6d6445a46/src/virtualship/instruments/base.py#L273) this means the depth dim is automatically set right for VirtualShip in this case but any local data is not validated/adjusted. Having the `"positive": "down"` or `"positive": "up"` metadata on local files allows the validation procedure to work in VirtualShip (in [`_get_local_ds()`](https://github.com/Parcels-code/virtualship/blob/26d6681b761722d436ab82e0babc97f6d6445a46/src/virtualship/instruments/base.py#L282)).

erikvansebille: OK, I thought so. In that case I would remove "by Parcels" in the sentence, as it may confuse readers to assume that it's Parcels that does the check

Nevertheless, I think it would be worth exploring if we can also add such a check in Parcels v4!
@VeckoTheGecko, so you know whether the `"positive"` (up or down) attribute is standard (or even required?) in the SGrid convention?

Issue #249, opened 2025-11-25
Realistic time for instrument deployment


ammedd:
Update https://virtualship.readthedocs.io/en/latest/user-guide/assignments/Research_proposal_intro.html
and the code with deployment times for different instruments. 

CTD
Add time to lower the vessels speed to zero and set the vessel to the appropriate direction against swell, current, wind, etc = 15 minutes
Add time to deploy the CTD, and to recover = 10 minutes (2*5 min)
Add time for the way down and up, usually at 1 m/s (winch capacity) = twice your maximum depth (in seconds)
Add time to speed up again  and set course to the next station = 10 minutes

Drifters and argo floats
It takes about 20 minutes to slow down and speed up the ship before and after the deployment. You may assume that any configurations that need to be made can be done while traveling to the next station.

j-atkins:
MFP now has an "Export Cruise Data" option which contains depth information for each waypoint, making this enhancement much simpler to implement.

PR #372, opened 2026-08-13
Timedeltas and arrival/departure ports from MFP export

### LINKED ISSUE #362: Add depature/arrival ports and time deltas to expedition init
[MFP](https://nioz.marinefacilitiesplanning.com/cruiselocationplanning#) now has an "Export Cruise Data" option, containing a "Travel Time to Next" column. This will be very helpful for initialising expeditions, whereby only a start datetime needs to be prescribed, and then rest of the waypoint timings can be auto completed. 

Plus, it will export the coordinates of the port of departure/arrival, a longstanding feature request for VirtualShip, enhancing realism.

The `virtualship init` logic should therefore be updated, with a bit of thought going to how we want users to prescribe their start date and handling the departure and arrival ports as waypoints which must fit in the scheduling but without any measurements.

j-atkins:
This PR uses the new exported fields from the MFP "Cruise Data" (see #362) to auto-fill the waypoints times using a user-prescribed `--start-date` and timedeltas from MFP when doing `virtualship init --from-mfp`.

Also, now that the ports of arrival/departure have been added to the MFP export these are now incorporated into the expedition.yaml, including a new `Port` model class. These act as special waypoints with times and locations but no instruments.

#### Further notes

- `Schedule` model gets pydantic field validators, ensuring that departure and arrival ports are always ingested in an expedition (dummies, or rather placeholders, are added when not specified in `virtualship init`). Ensures compliance with logic throughout codebase (e.g. with the RLCs).
    - These validators concern the structure of the `Schedule`, as a 'minimum standard' in order for other logic to work. Further checks (e.g. that waypoints can be reached in time) remain in `Schedule.verify` method. This is a design choice whereby these verification checks require more user thinking/interaction to relate to real-life.
    - Placeholder/dummy Port placeholders are eventually ignored in the simulations though, e.g. using new methods such as `Schedule._get_wps_in_use()`, so they won't be sailed to/from.
- Refactored `virtualship init` logic to a new module (cli/_initialise.py). Should help with readability and maintainability now that there is a lot more logic associated with this command, plus is consistent with the structure for other commands (`plan` and `run`).
- Now adds comments to the expedition.yaml, marking e.g. `# Port of Departure`, `Waypoint 1`, `Waypoint 2` and so on. This helps with config readability and is a previous user-feedback request.
- Some of the new logic relies on the MFP export format staying the same and we do not have control over this. I have added checks along the way when ingesting (with instructions to raise an issue if a user believes something has changed). Ideally though we would also unit test for this to catch drift, but I don't know how we would go about this given it requires getting an up to date export from MFP.
- Plan tool updated to handle Ports.
    - Ports are added in the "Waypoints" drop down but have no instrument selection options.
    - Adding/removing waypoints ignores Ports, these should not be removed/more added.

#### Separate PRs (to follow)

- Handle realistic instrument deployment timings (#249), the default 45 minutes added in MFP interferes with the stationkeeping time logic in simulate_schedule.
- Update docs.

-----

Closes #362


src/virtualship/cli/_initialise.py
```diff
@@ -0,0 +1,260 @@
...
+    path: str | Path, from_mfp: str | None = None, start_date: str | None = None
+):
+    path = Path(path)
+    path.mkdir(exist_ok=True)
+
+    expedition = path / EXPEDITION
+
+    if expedition.exists():
+        raise FileExistsError(
+            f"File '{expedition}' already exists. Please remove it or choose another directory."
```
erikvansebille: Is "choose" the best word here? Suggests an option selection/picker. Better to use "use"?

src/virtualship/cli/_initialise.py
```diff
@@ -0,0 +1,260 @@
...
+        if not has_departure:
+            dept_row = _create_port_row(expected_columns, "Departure Port")
+            mfp_data = pd.concat([dept_row, mfp_data], ignore_index=True)  # first row
+
+        if not has_arrival:
+            arr_row = _create_port_row(expected_columns, "Arrival Port")
+            mfp_data = pd.concat([mfp_data, arr_row], ignore_index=True)  # last row
+
+    # Drop unexpected columns
+    mfp_data = mfp_data[list(expected_columns)]
```
erikvansebille: Should his throw a warning too? Or will some columns always be dropped? If the latter, then perhaps "unexpected" should be replaced by "unnecessary"?

src/virtualship/models/expedition.py
```diff
@@ -70,6 +76,30 @@ def get_instruments(self) -> set[InstrumentType]:
...
+        raw = yaml.dump(self.model_dump(by_alias=True), default_flow_style=False)
+
+        lines = raw.splitlines(keepends=True)
+        annotated = []
+        waypoint_number = 0
+        for line in lines:
+            stripped = line.lstrip()
+            indent = " " * (len(line) - len(stripped))
+
+            # waypoints start with "- instrument:" and Ports start with "- location:" (no instrument field).
```
erikvansebille: Is this robust enough? What if users swap the order of fields in the yaml? Can that happen?

Would an alternative be to have an instrument "Port" or something?

src/virtualship/models/expedition.py
```diff
@@ -84,9 +114,34 @@ class ShipConfig(pydantic.BaseModel):
...
+    def _validate_waypoints(cls, value: list[Port | Waypoint]) -> list[Port | Waypoint]:
+        """Ensure first and last waypoints are Port objects and schedule contains non-port waypoints."""
+        if not isinstance(value[0], Port) or not isinstance(value[-1], Port):
+            raise ScheduleError(
+                "First and last waypoints must be Ports (of arrival/departure). "
+                "One or the other is currently missing."
+            )
+
+        if not any(isinstance(wp, Waypoint) for wp in value):
+            raise ScheduleError("At least one non-port waypoint must be provided.")
```
erikvansebille: Why is this? Will simulate otherwise break?

erikvansebille (commented):
Impressively heavy PR, but looks very good! A few small comments/suggestions below

src/virtualship/cli/_initialise.py
j-atkins: They would indeed be "unexpected" because a normal export should have exactly the `expected_columns`. I agree a warning would therefore be good. There is a warning to similar effect on L154-159, which I will adjust to explictly mention these columns will be dropped.

src/virtualship/models/expedition.py
j-atkins: It felt like it should be a requirement, but that being said perhaps it is good to still allow a simple Port to Port expedition.

erikvansebille: Yes, if there's no technical reason then I don't think it should be a requirement. Examples where it could be useful to have Port-to-Port schedules is as part of a larger multi-expedition plan where some of the time is simply travel between expeditions

src/virtualship/models/expedition.py
j-atkins: A user re-ordering of existing fields shouldn't matter, but I agree this implementation isn't robust enough if a _new_ field is added. If that new field comes before `- instrument` or `- location` then the logic breaks, which indeed feels too fragile. I'll address this.

src/virtualship/models/expedition.py
j-atkins: Now implemented in [9e9b2ff](https://github.com/Parcels-code/virtualship/pull/372/commits/9e9b2ff15e318fbbf3a6596a426042b505a54569)

erikvansebille (approved):
Looks good; happy to be merged

PR #388, opened 2026-09-17
Docs cruise data

j-atkins:
Building on #372, but keeping separate here to avoid adding more changes to what is already a large PR, this PR updates the docs with adjusted guidance/instructions for using Ports in the VirtualShip workflow. Also clarifies that datetimes are auto-filled in the expedition.yaml after providing a `--start-date` in `virtualship init`.


erikvansebille (approved):
Looks good, no comments

